PR #69, opened 2015-01-21
Replaced uint with usize and the 'u' type suffix with 'us'.

### LINKED ISSUE #67: Replace int/uint with isize/usize or explicit sizes

mmatyas:
A simple warning fix, resolves #67 .

### DIFF
--- a/src/util/str.rs
+++ b/src/util/str.rs
@@ -97,12 +97,12 @@ impl Ascii {
 
     #[inline]
     pub fn to_lowercase(self) -> Ascii {
-        Ascii { chr: ASCII_LOWER_MAP[self.chr as uint] }
+        Ascii { chr: ASCII_LOWER_MAP[self.chr as usize] }
     }
 
     #[inline]
     pub fn eq_ignore_case(self, other: Ascii) -> bool {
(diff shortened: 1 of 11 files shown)


aroben:
I think this got taken care of in #79

kmcallister:
Not all of it.

mmatyas:
Most of the patch is still valid, so I updated and rebased it to master.

kmcallister:
Thanks. Did you think about whether some of these should be fixed-size integers instead? I haven't gone through and thought about that yet -- that's basically my remaining review step.

src/util/str.rs
```diff
@@ -112,7 +112,7 @@ pub trait AsciiCast {
 
 impl AsciiCast for char {
     fn to_ascii_opt(&self) -> Option<Ascii> {
-        let n = *self as uint;
+        let n = *self as usize;
```
kmcallister: This should be `u32` I guess, because it comes from a `char`.  I'm not sure why it was `uint`, but I think I copied this code from libstd.

kmcallister:
It does seem like `usize` is appropriate in these cases, with the one exception that I noted above.

mmatyas:
Ok, patch updated.

kmcallister:
Tested locally and merged :)

PR #71, opened 2015-01-27
Fix the build on 2015-01-25 Rust nightly

aroben:
This fixes all build errors I was seeing in the 2015-01-25 Rust nightly build. `make examples check bench` now runs without errors.


kmcallister:
Thanks!

PR #83, opened 2015-02-02
Shrink test output in CI

aroben:
Travis CI won't display build logs longer than 10,000 lines. Since we have so many tests, our CI logs get truncated and you can't see the test failures down at the end of the log.

Instead of printing each test on its own line, we now just print out a single character for each test indiciating success (`.`), failure (`F`), or that the test was ignored (`I`). Failing tests are still printed in full at the end of the test run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ script:
   - mkdir build
   - cd build
   - ../configure
-  - make check docs for_c
+  - make check docs for_c | ../scripts/shrink-test-output.py
 after_script:
   - cd /home/travis/build/servo/html5ever/target
   - curl http://www.rust-ci.org/artifacts/put?t=$RUSTCI_TOKEN | sh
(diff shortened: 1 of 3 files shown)


aroben:
Here's an example build with a failing test: https://travis-ci.org/servo/html5ever/builds/49209191

And an example build with all passing (or ignored) tests: https://travis-ci.org/servo/html5ever/builds/49209313

kmcallister:
Nice trick :)

Can you move the output condenser script to `scripts/` so it's easier to use outside of Travis?  (Also, I'd prefer Python over Perl, but either is fine really.)

aroben:
OK, I moved it to a separate Python script.

aroben:
Uh oh, the build "succeeded" even though compiling failed. Guess we need to get the exit status of `make` rather than the exit status of the script.

aroben:
The exit code issue is fixed. Let me know what you think. Maybe this is too much complexity for the benefit?

kmcallister:
Looks great, thanks :)

PR #90, opened 2015-02-09
Fix compilation/warnings with rustc 00df3251f 2015-02-08

aroben:


zmbush:
Any word on this?
 It would be great to get it merged.

PR #105, opened 2015-02-28
Use enum rather than bool for controlling serialize

ChrisParis:
In https://github.com/servo/servo/pull/5029, @jdm [requested](https://critic.hoppipolla.co.uk/showcomment?chain=10737) an enum rather than bool for indicating whether to serialize a whole node, or only its children. For consistency, I pushed that suggestion all the way down through html5ever, and it looks like this.


kmcallister:
Could you please rebase this so Travis will test it with the latest rustup?

PR #107, opened 2015-03-11
Add a quiescent state to the tokenizer to allow interrupting the parser ...

jdm:
...and resuming it later.

### DIFF
--- a/src/tokenizer/interface.rs
+++ b/src/tokenizer/interface.rs
@@ -108,7 +108,7 @@ pub trait TokenSink {
     /// Process a token.
     fn process_token(&mut self, token: Token);
 
-    /// The tokenizer will call this after emitting any start tag.
+    /// The tokenizer will call this after emitting any tag.
     /// This allows the tree builder to change the tokenizer's state.
     /// By default no state changes occur.
     fn query_state_change(&mut self) -> Option<states::State> {
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -328,8 +328,8 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         }
     }
 
-    // Run the state machine for as long as we can.
-    fn run(&mut self) {
+    /// Run the state machine for as long as we can.
+    pub fn run(&mut self) {
         if self.opts.profile {
             loop {
                 let state = self.state;
@@ -410,11 +410,9 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         });
         self.process_token(token);
 
-        if self.current_tag_kind == StartTag {
-            match self.sink.query_state_change() {
-                None => (),
-                Some(s) => self.state = s,
-            }
+        match self.sink.query_state_change() {
+            None => (),
+            Some(s) => self.state = s,
         }
     }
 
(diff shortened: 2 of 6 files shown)


src/tokenizer/mod.rs
```diff
@@ -227,18 +227,21 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+            } else {
+                0
+            };
+
+            self.input_buffers.push_back(input, pos);
+        }
 
-        self.input_buffers.push_back(input, pos);
         self.run();
     }
```
kmcallister: Are these changes basically to allow calling `feed` with an empty string to re-start parsing? I'd rather just make `run` public.

src/tokenizer/mod.rs
```diff
@@ -410,11 +413,9 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
         self.process_token(token);
 
-        if self.current_tag_kind == StartTag {
-            match self.sink.query_state_change() {
-                None => (),
-                Some(s) => self.state = s,
-            }
+        match self.sink.query_state_change() {
+            None => (),
+            Some(s) => self.state = s,
```
kmcallister: You should change the docs in `tokenizer/interface.rs` to match.

src/tokenizer/mod.rs
```diff
@@ -651,6 +652,11 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
         h5e_debug!("processing in state {:?}", self.state);
         match self.state {
+            states::Quiescent => {
+                self.state = states::Data;
+                return false;
+            }
```
kmcallister: This could use a comment along the lines of

> Reachable only through `query_state_change`. The tree builder wants the tokenizer to suspend processing.

src/tree_builder/interface.rs
```diff
@@ -40,6 +40,15 @@ pub enum NodeOrText<Handle> {
     AppendText(String),
 }
 
+/// Whether to interrupt further parsing of the current input until
+/// the next explicit resumption of the tokenizer, or continue without
+/// any interruption.
+#[derive(PartialEq)]
+pub enum NextParserState {
+    Suspend,
+    Continue
```
kmcallister: Nit: should have a trailing comma.

src/tree_builder/interface.rs
```diff
@@ -40,6 +40,15 @@ pub enum NodeOrText<Handle> {
     AppendText(String),
 }
 
+/// Whether to interrupt further parsing of the current input until
+/// the next explicit resumption of the tokenizer, or continue without
+/// any interruption.
+#[derive(PartialEq)]
```
kmcallister: Should derive the same set of traits as `QuirksMode` above.

src/tokenizer/mod.rs
jdm: Why? It makes the code in Servo easier, since it doesn't have to special-case restarting parsing when no input is available vs. feeding a new block of input that has been received.

kmcallister: You can call `run` and then, if there's input available, call `feed`. Seems logical enough.

kmcallister:
@jdm: Any update on this?

jdm:
Changes pushed.

src/tokenizer/interface.rs
```diff
@@ -110,7 +110,8 @@ pub trait TokenSink {
 
     /// The tokenizer will call this after emitting any start tag.
     /// This allows the tree builder to change the tokenizer's state.
-    /// By default no state changes occur.
+    /// By default a <script> element will put the tokenizer into a
+    /// quiescent state, but no other changes occur.
```
kmcallister: Isn't the default return value `NextParserState::Continue`?

jdm: That's true. I guess I'm not sure what documentation changes you're looking for here.

kmcallister: It's now called for all tags, not just start tags.

jdm:
Change pushed.

jdm:
Fixed.

kmcallister:
r=me with squash

jdm:
Squashed.

PR #111, opened 2015-03-17
Upgrade to rustc 1.0.0-dev (1760e8749 2015-03-16) (built 2015-03-17).

Ogeon:
An attempt to upgrade to the latest rust version. The library builds, but the tests fails with two ICEs. Probably [#23110](https://github.com/rust-lang/rust/issues/23110) and [#23084](https://github.com/rust-lang/rust/issues/23084). I decided to submit it anyway, in case it becomes useful.

### DIFF
--- a/examples/html2html.rs
+++ b/examples/html2html.rs
@@ -7,8 +7,6 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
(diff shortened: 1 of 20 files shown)


kmcallister:
Merged, thanks!

PR #113, opened 2015-03-21
Upgrade to rustc 1.0.0-dev (68d694156 2015-03-20)

o01eg:
Fixed using features.


Ogeon:
This change is also required by the nightlies now.

PR #114, opened 2015-03-24
Implement zero-copy parsing

kmcallister:
This is based on #60 but with substantial changes. The biggest difference is that we only use shared buffers for the character runs found by `pop_except_from`. The majority of the remaining spans are single ASCII characters, which have their own fast path. Everything else is a `String` as before.

This branch also drops many of the micro-optimizations from #60. Unlike that PR, we leave the parser rules alone for the most part.

r? @Manishearth or @SimonSapin (general review)

r? @cgaebel (iobuf usage in `tendril.rs`)

Depending on the specific content and the I/O chunk size, this branch speeds up tokenization by up to a few percent. I did not see any significant performance regressions with sensible chunk sizes.

I have plans for further optimizations, including following up on the rustc bugs @cgaebel identified in #60.

The branch already achieves a significant drop in allocations and memory consumption:

(preliminary numbers)

### Webapp spec, single page:

pre-zerocopy

```
846,520 allocs, 37,567,992 bytes allocated
maximum resident: 11,248 kB
```

zerocopy

```
95,690 allocs, 25,698,480 bytes allocated
maximum resident: 3,648 kB
```

### Wikipedia (GotG from servo-static-suite)

pre-zerocopy

```
62,705 allocs, 2,299,424 bytes allocated
maximum resident: 3,948 kB
```

zerocopy

```
11,549 allocs, 2,355,984 bytes allocated
maximum resident: 3,596 kB
```

### DIFF
--- a/src/tokenizer/buffer_queue.rs
+++ b/src/tokenizer/buffer_queue.rs
@@ -10,6 +10,7 @@
 use core::prelude::*;
 
 use util::str::AsciiCast;
+use util::tendril::Tendril;
 use util::smallcharset::SmallCharSet;
 
 use core::str::CharRange;
@@ -20,16 +21,16 @@ pub use self::SetResult::{FromSet, NotFromSet};
 
 struct Buffer {
     /// Byte position within the buffer.
-    pub pos: usize,
+    pub pos: u32,
     /// The buffer.
-    pub buf: String,
+    pub buf: Tendril,
 }
 
 /// Result from `pop_except_from`.
 #[derive(PartialEq, Eq, Debug)]
 pub enum SetResult {
     FromSet(char),
-    NotFromSet(String),
+    NotFromSet(Tendril),
 }
 
 /// A queue of owned string buffers, which supports incrementally
@@ -43,12 +44,12 @@ impl BufferQueue {
     /// Create an empty BufferQueue.
     pub fn new() -> BufferQueue {
         BufferQueue {
-            buffers: VecDeque::with_capacity(3),
+            buffers: VecDeque::with_capacity(16),
         }
     }
 
     /// Add a buffer to the beginning of the queue.
-    pub fn push_front(&mut self, buf: String) {
+    pub fn push_front(&mut self, buf: Tendril) {
         if buf.len() == 0 {
             return;
         }
@@ -61,8 +62,8 @@ impl BufferQueue {
     /// Add a buffer to the end of the queue.
     /// 'pos' can be non-zero to remove that many bytes
     /// from the beginning.
-    pub fn push_back(&mut self, buf: String, pos: usize) {
-        if pos >= buf.len() {
+    pub fn push_back(&mut self, buf: Tendril, pos: u32) {
+        if pos as usize >= buf.len() {
             return;
         }
         self.buffers.push_back(Buffer {
@@ -74,7 +75,7 @@ impl BufferQueue {
     /// Look at the next available character, if any.
     pub fn peek(&mut self) -> Option<char> {
         match self.buffers.front() {
-            Some(&Buffer { pos, ref buf }) => Some(buf.as_slice().char_at(pos)),
+            Some(&Buffer { pos, ref buf }) => Some(buf.as_slice().char_at(pos as usize)),
             None => None,
         }
     }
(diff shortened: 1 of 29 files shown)


src/tokenizer/buffer_queue.rs
```diff
@@ -20,16 +21,16 @@ pub use self::SetResult::{FromSet, NotFromSet};
 
 struct Buffer {
     /// Byte position within the buffer.
-    pub pos: usize,
+    pub pos: u32,
     /// The buffer.
-    pub buf: String,
+    pub buf: Tendril,
```
cgaebel: Isn't `pos` redundant with `buf` now, since Iobuf has its own cursor?

src/tokenizer/char_ref/mod.rs
```diff
@@ -204,7 +205,7 @@ impl CharRefTokenizer {
     }
 
     fn unconsume_numeric<Sink: TokenSink>(&mut self, tokenizer: &mut Tokenizer<Sink>) -> Status {
-        let mut unconsume = String::from_str("#");
+        let mut unconsume = Tendril::from_char('#');
```
cgaebel: Maybe `from_u8(b'#')` would be a better pattern to use here...

src/util/tendril.rs
```diff
@@ -0,0 +1,922 @@
...
+}
+
+impl Deref for Tendril {
+    type Target = str;
+
+    #[inline]
+    fn deref<'a>(&'a self) -> &'a str {
+        match self.0 {
+            Shared(ref s) => unsafe {
+                mem::transmute(s.as_window_slice())
```
cgaebel: Can't this also use `str::from_utf8_unchecked`?

src/util/tendril.rs
```diff
@@ -0,0 +1,922 @@
...
+        match self.0 {
+            Ascii(b'\n') => *self = Tendril::new(),
+            Ascii(_) => (),
+            Owned(ref mut s) => {
+                if s.starts_with("\n") {
+                    s.remove(0);
+                }
+            }
+            Shared(ref mut b) => unsafe {
+                if b.unsafe_peek_le(0) == b'\n' {
```
cgaebel: Might want a `debug_assert!(b.len() >= 1)` here, just to be safe.

cgaebel:
Doesn't this mean that any long string of text (i.e. that spans multiple chunks) would have to fall back to a `String`? How big are chunks in Servo?

src/util/tendril.rs
```diff
@@ -0,0 +1,922 @@
...
+        self.0 = Owned(String::from_str(self));
+        match self.0 {
+            Owned(ref mut s) => s,
+            _ => unreachable!(),
+        }
+    }
+
+    #[inline(always)]
+    fn check_len(&self) {
+        if self.len() > TENDRIL_MAX_LEN as usize {
```
cgaebel: Why not just use `len32`?

src/util/tendril.rs
```diff
@@ -0,0 +1,922 @@
...
+        // Try to merge adjacent Iobufs.
+        if let (&mut Tendril(Shared(ref mut a)), &Tendril(Shared(ref b)))
+            = (&mut *self, &rhs)
+        {
+            if a.extend_with(b).is_ok() {
+                return;
+            }
+        }
+
+        if rhs.len() > TENDRIL_MAX_LEN as usize{
```
cgaebel: `len32`?

src/util/tendril.rs
```diff
@@ -0,0 +1,922 @@
...
+
+        // Try to merge adjacent Iobufs.
+        if let (&mut Tendril(Shared(ref mut a)), &Tendril(Shared(ref b)))
+            = (&mut *self, &rhs)
+        {
+            if a.extend_with(b).is_ok() {
+                return;
+            }
+        }
+
```
cgaebel: I think everything below this point should be out in a function marked `#[cold]`.

src/util/tendril.rs
```diff
@@ -0,0 +1,922 @@
...
+                if let Some(n) = as_single_ascii(rhs) {
+                    self.0 = Ascii(n);
+                    return;
+                }
+            }
+            n if n > TENDRIL_MAX_LEN as usize => {
+                panic!("attempted to extend tendril by more than 1 GB");
+            }
+
+            // Otherwise, 2 * TENDRIL_MAX_LEN does not overflow u32.
```
cgaebel: One needs to worry about overflowing `i32`, not `u32`. I use a bit in one of the `u32` fields in iobuf to say whether or not the pointer is owned or borrowed.

cgaebel:
Iobuf usage looks good to me (modulo comments)! I like this remix.

src/tokenizer/buffer_queue.rs
kmcallister: But the owned variant of a `Tendril` doesn't.

Eventually I would like to replace `BufferQueue` with a rope, as discussed elsewhere.

cgaebel: Good point!

kmcallister:
Now #141.

PR #116, opened 2015-03-28
Upgrade to rustc 1.0.0-nightly (c89de2c56 2015-03-28) (built 2015-03-29)

Ogeon:
Removed a few deprecated things, such as a bunch of `as_slice`. I decided to let most of the `char_at` and friends stay for a little longer.

The tests are, once again, experiencing the [#23110](https://github.com/rust-lang/rust/issues/23110) ICE and I'm not sure where it originates this time.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -11,7 +11,7 @@
 #![crate_type="dylib"]
 
 #![feature(plugin_registrar, quote)]
-#![feature(rustc_private, convert)]
+#![feature(rustc_private, convert, slice_patterns)]
 #![deny(warnings)]
 
 extern crate syntax;
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -11,12 +11,12 @@
 #![crate_type="dylib"]
 
 #![feature(plugin_registrar, quote)]
-#![feature(rustc_private, core, std_misc)]
+#![feature(rustc_private, convert)]
 #![deny(warnings)]
 
 extern crate syntax;
 extern crate rustc;
-extern crate "rustc-serialize" as rustc_serialize;
+extern crate rustc_serialize;
 
 #[macro_use]
 extern crate mac;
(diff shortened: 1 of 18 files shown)


Manishearth:
Travis is [ICEing](https://travis-ci.org/servo/html5ever/builds/56209902) :/

/me tries locally

Ogeon:
Yes, as I said. I have the same ICE in my own project where I'm using html5ever. I have no Idea where it comes from. There are no multi line `#![feature(...)]`.

Manishearth:
Oh, right. Hm

macros/src/lib.rs
```diff
@@ -11,12 +11,12 @@
 #![crate_type="dylib"]
 
 #![feature(plugin_registrar, quote)]
-#![feature(rustc_private, core, std_misc)]
+#![feature(rustc_private, convert)]
```
Ygg01: Hm, is this example of multi-line feature?

Ogeon: I tried to change them. No difference. Besides, they are in the macro crate, which is built before the html5ever crate. I can still change it if it's desired.

Ygg01: @Ogeon, I don't think the issue is necessarily with `#![feature]`. I've been hitting a similar bug, without using `#![feature]` at all and on older version of Rust. I suspect it's some underlying macro expansion error.

Ogeon: Sounds reasonable.

Ogeon:
This is really strange... My project builds now, without any ICE, when using these fixes, but these tests are still failing. Nothing changed in my code.

Ygg01:
@Ogeon : Did you run the `make check examples `. That's what causes ICE for me :cry:

Ogeon:
I did run `make check` and it failed, but I could successfully build [symbiosis](https://github.com/Ogeon/symbiosis), where I'm using html5ever. That's what's odd.

kmcallister:
I think all these changes are on master now. I worked around the ICE building examples by disabling debug info.

PR #117, opened 2015-04-03
Rust upgrade to rustc 1.0.0-nightly (d17d6e7f1 2015-04-02)

Ygg01:
Rustup

rustc 1.0.0-nightly (d17d6e7f1 2015-04-02) (built 2015-04-03)

Minor note: make for_c doesn't work. Not sure why.

### DIFF
--- a/examples/tokenize.rs
+++ b/examples/tokenize.rs
@@ -18,7 +18,7 @@ use html5ever::tokenizer::{TokenSink, Token, TokenizerOpts, ParseError};
 use html5ever::tokenizer::{CharacterTokens, NullCharacterToken, TagToken, StartTag, EndTag};
 use html5ever::driver::{tokenize_to, one_input};
 
-#[derive(Copy)]
+#[derive(Copy, Clone)]
 struct TokenPrinter {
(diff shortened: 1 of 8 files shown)


kmcallister:
> Minor note: make for_c doesn't work. Not sure why.

We can't merge the patch with that broken. But you seem to have fixed the problem, going by Travis.

Ygg01:
@kmcallister Initially, my make script misbehaved (most likely due to Rust upgrade), which looked like some random error (something about multiple filenames). I didn't saw what the error was, until Travis CI threw it and then after `cargo clean && rm -r /build` it started displaying errors correctly.

As far as braces are concerned, when I was cherry-picking I seemed to have missed this edit and had to throw it at last minute, forgetting braces. My apologizes.

Ah, so it was the correct solution :relieved: . I was feeling a bit worried, because I'm uncertain if `make for_c` does any testing.

kmcallister:
Thanks!

> I'm uncertain if `make for_c` does any testing

It doesn't. :) The C API is incomplete and I haven't written tests yet.

PR #124, opened 2015-04-13
Upgrade to rustc 1.0.0-nightly (6790b0e51 2015-04-11) (built 2015-04-12)

SimonSapin:
Also disable doctests because of https://github.com/rust-lang/cargo/issues/1512

Note that there weren’t any doctest to run anyway.

r? @kmcallister

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,13 +4,17 @@ name = "html5ever"
 version = "0.0.0"
 authors = [ "The html5ever Project Developers" ]
 
+[lib]
+name = "html5ever"
+
+# https://github.com/rust-lang/cargo/issues/1512
+doctest = false
+
 [dependencies]
 time = "0"
 log = "0"
-
-# servo/string-cache#80
-phf = "0.6"
-phf_macros = "0.6"
+phf = "0.7"
+phf_macros = "0.7"
 
 [dependencies.string_cache]
 git = "https://github.com/servo/string-cache"
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -22,6 +22,16 @@ extern crate mac;
 
 use rustc::plugin::Registry;
 
+macro_rules! panictry {
+    ($e:expr) => ({
+        use syntax::diagnostic::FatalError;
+        match $e {
+            Ok(e) => e,
+            Err(FatalError) => panic!(FatalError)
+        }
+    })
+}
+
 // Make these public so that rustdoc will generate documentation for them.
 pub mod named_entities;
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -4,13 +4,17 @@ name = "html5ever"
...
+
 [dependencies]
 time = "0"
 log = "0"
-
-# servo/string-cache#80
-phf = "0.6"
-phf_macros = "0.6"
+phf = "0.7"
+phf_macros = "0.7"
```
kmcallister: These should go back to `= "0"` unless there's a reason to keep them pinned.

macros/src/lib.rs
```diff
@@ -22,6 +22,16 @@ extern crate mac;
...
+macro_rules! panictry {
+    ($e:expr) => ({
+        use syntax::diagnostic::FatalError;
+        match $e {
+            Ok(e) => e,
+            Err(FatalError) => panic!(FatalError)
+        }
+    })
+}
+
```
kmcallister: Why do we need this instead of `Result::unwrap()`?

SimonSapin: This is what libsyntax does:

https://github.com/rust-lang/rust/blob/b9ed9e2a326cafa6f2445aec7221dc419089cf5f/src/libsyntax/lib.rs#L53-L56
> A variant of 'try!' that panics on Err(FatalError). This is used as a crutch on the way towards a non-panic!-prone parser. It should be used for fatal parsing errors; eventually we plan to convert all code using panictry to just use normal try

Right now it’s the same as `.unwrap`, but the plan is apparently that it should be `try!` eventually.

Cargo.toml
SimonSapin: Semver says anything goes before 1.0, but Cargo’s extrapolation of it is that any 0.7.x should be compatible with 0.7.y with y <= x, while 0.8 is not. So if phf follows that convention, 0.8 will probably break html5ever while (in a post-Rust-1.0 world) 0.7 will still be available and compatible.

That said, I’ll make them `= "0"` if you still feel it’s the right thing to do.

kmcallister: Okay, `= "0.7"` works for me.

macros/src/lib.rs
kmcallister: Ah, can you add a link or reference to that libsyntax code in a comment here? And, is there an associated Rust issue number?

kmcallister:
Thanks! r=me with the link or reference to libsyntax as mentioned above.

PR #128, opened 2015-04-25
Make the `sink` module an optional feature.

mbrubeck:
Consumers who provide their own sinks (like Servo) probably don't need this module.  Disabling it reduces compile time by about 20% (a couple of seconds on my laptop).  Probably reduces final link time and possibly binary size too, but I haven't measured those.

r? @kmcallister

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,9 @@ rustc-serialize = "0"
 [dev-dependencies.test_util]
 path = "test_util"
 
+[dev-dependencies.html5ever_dom_sink]
+path = "dom_sink"
+
 [profile.dev]
 debug = false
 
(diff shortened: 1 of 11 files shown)


kmcallister:
Maybe they should move to an external crate, even a separate repo. We just need that crate as a dev-dependency to build the examples and test suite.

Ygg01:
:+1:  

Modularize all the things

mbrubeck:
Updated to move `html5ever::sink` into a separate `html5ever_dom_sink` crate.  Bikeshedding of the crate name is welcome.

Besides moving the code, the only other change was turning `RcDom::Handle` into a newtype struct, to satisfy the coherence rules on `impl Serializable for Handle`.

kmcallister:
Merged, thanks!

PR #137, opened 2015-05-26
Implement support for form owner

mukilan:
This is needed for servo/servo#3553.

[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/137)

### DIFF
--- a/dom_sink/src/owned_dom.rs
+++ b/dom_sink/src/owned_dom.rs
@@ -197,6 +197,17 @@ impl TreeSink for Sink {
         x == y
     }
 
+    fn same_home_subtree(&self, _x: Handle, _y: Handle) -> bool {
+        true
+    }
+
+    fn associate_with_form(&mut self, _target: Handle, _form: Handle) {
+    }
+
+    fn has_parent_node(&self, node: Handle) -> bool {
+        !node.parent.is_null()
+    }
+
     fn elem_name(&self, target: Handle) -> QualName {
         match target.node {
             Element(ref name, _) => name.clone(),
@@ -230,8 +241,9 @@ impl TreeSink for Sink {
 
     fn append_before_sibling(&mut self,
             sibling: Handle,
-            child: NodeOrText<Handle>) -> Result<(), NodeOrText<Handle>> {
-        let (mut parent, i) = unwrap_or_return!(get_parent_and_index(sibling), Err(child));
+            child: NodeOrText<Handle>) {
+        let (mut parent, i) = get_parent_and_index(sibling)
+            .expect("append_before_sibling called on node without parent");
 
         let mut child = match (child, i) {
             // No previous node.
@@ -241,7 +253,7 @@ impl TreeSink for Sink {
             (AppendText(text), i) => {
                 let prev = parent.children[i-1];
                 if append_to_existing_text(prev, &text) {
-                    return Ok(());
+                    return;
                 }
                 self.new_node(Text(text))
             }
@@ -259,7 +271,6 @@ impl TreeSink for Sink {
 
         child.parent = parent;
         parent.children.insert(i, child);
-        Ok(())
     }
 
     fn append_doctype_to_document(&mut self, name: String, public_id: String, system_id: String) {
--- a/src/tree_builder/actions.rs
+++ b/src/tree_builder/actions.rs
@@ -207,40 +207,8 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
 
     // Insert at the "appropriate place for inserting a node".
     fn insert_appropriately(&mut self, child: NodeOrText<Handle>, override_target: Option<Handle>) {
-        declare_tag_set!(foster_target = table tbody tfoot thead tr);
-        let target = override_target.unwrap_or_else(|| self.current_node());
-        if !(self.foster_parenting && self.elem_in(target.clone(), foster_target)) {
-            // No foster parenting (the common case).
-            return self.sink.append(target, child);
-        }
-
-        // Foster parenting
-        // FIXME: <template>
-        let last_table = self.open_elems.iter()
-            .enumerate()
-            .rev()
-            .filter(|&(_, e)| self.html_elem_named(e.clone(), atom!(table)))
-            .next();
-
-        match last_table {
-            None => {
-                let html_elem = self.html_elem();
-                self.sink.append(html_elem, child);
-            }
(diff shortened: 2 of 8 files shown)


Manishearth:
\o/

ITS HAPPENING!

dom_sink/src/owned_dom.rs
```diff
@@ -230,8 +241,9 @@ impl TreeSink for Sink {
 
     fn append_before_sibling(&mut self,
             sibling: Handle,
-            child: NodeOrText<Handle>) -> Result<(), NodeOrText<Handle>> {
-        let (mut parent, i) = unwrap_or_return!(get_parent_and_index(sibling), Err(child));
```
Manishearth: Perhaps we should avoid crashing here or document the panicky behavior?

Manishearth:
r? @kmcallister 

(I'll go through this anyway, but I'm not nearly familiar with parsing to sign off on it)

dom_sink/src/owned_dom.rs
```diff
@@ -197,6 +197,17 @@ impl TreeSink for Sink {
         x == y
     }
 
+    fn same_home_subtree(&self, _x: Handle, _y: Handle) -> bool {
+        true
+    }
+
+    fn associate_with_form(&mut self, _target: Handle, _form: Handle) {
+    }
```
SimonSapin: Could this (and `same_home_subtree`?) be a default implementation in the trait?

src/tree_builder/interface.rs
```diff
@@ -61,6 +61,10 @@ pub trait TreeSink {
     /// Do two handles refer to the same node?
     fn same_node(&self, x: Self::Handle, y: Self::Handle) -> bool;
 
+    /// Are two handles present in the same tree
+    /// https://html.spec.whatwg.org/multipage/infrastructure.html#home-subtree
```
SimonSapin: I don’t understand how this can ever be false. Could the doc-comment give an example?

mukilan: As far as I understand there are two cases where it will be false:

1) Fragment Parsing case i.e ```elem.innerHTML =  <input type="text" />``` . In this case the parser won't call ```associate_form_owner``` even when one of ```elem```'s ancestor is a form element . But the form owner will be set when the DocumentFragment's children are inserted into the context element, after parsing.

2) An inline script element removes the form element from the Document before the ```<input>``` is parsed and ```form_elem``` points to a form element that is not an ancestor of the input tag (the wierd table case)

I must admit that I don't fully understand case 1. In particular, why does the fragment parser algorithm initialize the parser's form element pointer to the closest form element ancestor of the context element? (https://html.spec.whatwg.org/multipage/syntax.html#parsing-html-fragments:form-element-pointer step 11)

I'm not sure if there are other cases that I might have missed.

dom_sink/src/owned_dom.rs
mukilan: I did this because I thought owned_dom and rc_dom were just examples and don't implement all features (form owner support in this case). For real clients wouldn't we want to enforce the implementation of form owners? Or is it optional?

dom_sink/src/owned_dom.rs
mukilan: The comment in the trait definition documents the fact that append_before_sibling will only be called when has_parent(sibling) is true and hence it should never fail. Is this not enough?

Manishearth:
Travis fails, btw.

mukilan:
Fixed build errors.

metajack:
@mukilan Did you address @SimonSapin comments as well?

dom_sink/src/owned_dom.rs
SimonSapin: It can be useful to parse HTML in a "real" application that is not a browser doesn't support form submission.

jdm:
I agreed to look over these changes.

dom_sink/src/owned_dom.rs
mukilan: That makes sense. I'll change them to be default implementations.

src/tree_builder/mod.rs
```diff
@@ -364,6 +364,48 @@ impl<Handle, Sink> TreeBuilder<Handle, Sink>
...
+            .next();
+
+        match last_table {
+            None => {
+                LastChild(self.html_elem())
+            }
+            Some((idx, last_table)) => {
+                // Try inserting "inside last table's parent node, immediately before last table"
+                if self.sink.has_parent_node(last_table.clone()) {
+                    BeforeSibling(last_table.clone())
```
jdm: We've lost the fallback behaviour from https://github.com/servo/html5ever/pull/137/files#diff-3f339094e126866dfc7763b9a1d54beaL232, haven't we?

jdm: Nevermind, I've read the other comments and I understand why this change is appropriate.

src/tree_builder/actions.rs
```diff
@@ -755,10 +723,39 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
...
+        // Step 4.
+        // TODO: Handle template element case
+        if form_associatable(qname.clone())
+           && self.form_elem.is_some()
+           && !(reassociatable(qname.clone())
+               && attrs.iter().any(|a| a.name == qualname!("","form"))) {
+
+               let form = self.form_elem.as_ref().unwrap().clone();
+               if self.sink.same_home_subtree(tree_node, form.clone()) {
+                   self.sink.associate_with_form(elem.clone(), form)
```
mukilan: Hi @hsivonen, @jdm referred me to you for this particular query: is there a specific reason why the spec requires "associating the form owner" to happen before the insertion of the form control element into the tree? [Spec](https://html.spec.whatwg.org/multipage/syntax.html#creating-and-inserting-nodes:form-associated-element)

hsivonen:
@mukilan, I'm not sure. It seems to me that @Hixie found it more convenient to handle parser-specific stuff in the parser's "create an element for a token" and to stipulate an exception to the insertion algorithm than to define parser-specific stuff in the insertion algorithm itself.

There could be another reason that I fail to see, though.

nox:
@mukilan Is there really no way to keep `insert_appropriately()` as it is currently? I liked that it relied on inserting failing if there are no parent, instead of needing a `has_parent_node()` hook.

canaltinova:
We can close this since it's implemented in #249

PR #141, opened 2015-06-10
Work in progress -- zero copy parsing take 3

kmcallister:
This includes the [tendril](https://github.com/kmcallister/tendril) integration which is desired for `document.write`. The only missing piece afaik is fixing up the C API (or temporarily disabling it, but I'd rather not).

I won't have time to finish up this PR but I can answer questions for sure.


SimonSapin:
@kmcallister Could you remake that PR from a branch in this repository rather than in your fork? That way if someone else finishes it we can keep all the review/discussion in the same place. Thanks!

SimonSapin:
(I can do it if you prefer.)

PR #145, opened 2015-06-16
add documentation link to cargo.toml

kstep:


Ygg01:
Wouldn't it be better that documentation is hosted on servo related accounts?

SimonSapin:
We might do that at some point, but as long as @kmcallister’s setup works I don’t mind leaving docs there.

Thanks @kstep !

PR #153, opened 2015-08-18
Implement <template>

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/153)

### DIFF
--- a/src/tree_builder/rules.rs
+++ b/src/tree_builder/rules.rs
@@ -165,7 +165,7 @@ impl<Handle, Sink> TreeBuilderStep
                 }
 
                 tag @ </template> => {
-                    if !self.in_scope_named(default_scope, atom!(template)) {
+                    if !self.in_html_elem_named(atom!(template)) {
                         self.unexpected(&tag);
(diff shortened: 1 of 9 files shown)


src/tree_builder/rules.rs
```diff
@@ -1179,18 +1193,60 @@ impl<Handle, Sink> TreeBuilderStep
...
-                // extra logic at other points noted throughout the
-                // parser.
-
-                if self.opts.ignore_missing_rules {
-                    self.step(InBody, token)
-                } else {
-                    panic!("FIXME: <template> not implemented");
+            InTemplate => match_token!(token {
+                CharacterTokens(_, _) => self.step(InBody, token),
+                CommentToken(_) => self.step(InBody, token),
```
SimonSapin: Is `DocTypeToken` missing here?

nox: `DocTypeToken` is already implicitly rejected in all insertion modes (see for example the BeforeHtml branch), I just continued that trend.

PR #155, opened 2015-08-18
Improve add_attrs_if_missing()

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/155)

### DIFF
--- a/src/rcdom.rs
+++ b/src/rcdom.rs
@@ -13,6 +13,7 @@
 //! web browser using it. :)
 
 use std::cell::RefCell;
+use std::collections::HashSet;
 use std::default::Default;
 use std::borrow::Cow;
 use std::io::{self, Write};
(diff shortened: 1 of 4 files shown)


src/rcdom.rs
```diff
@@ -244,15 +245,14 @@ impl TreeSink for RcDom {
...
+        let existing = if let Element(_, ref mut attrs) = node.deref_mut().node {
+            attrs
+        } else {
+            panic!("not an element")
         };
 
-        // FIXME: quadratic time
-        attrs.retain(|attr|
-            !existing.iter().any(|e| e.name == attr.name));
+        let names = attrs.iter().map(|e| e.name.clone()).collect::<HashSet<_>>();
```
SimonSapin: `existing.iter()`, right?

Maybe rename `names` to `existing_names`, and move the `.retain()` call to a `.map()` on the result of `.into_iter()`

nox: I fixed that, but forgot to commit before pushing… Will address the other points while at it.

PR #156, opened 2015-08-18
Implement tokenization of CDATA sections

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/156)


nox:
@SimonSapin I made eat() take an Fn.

PR #162, opened 2015-08-19
Properly handle U+0000 NULL when dispatching tree construction

nox:
[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/162)


nox:
Fixed. Shouldn't be bit-rotted by previous PRs.

PR #168, opened 2015-09-01
Run tree builder and tokenizer tests again.

SimonSapin:
Something change in rustc and the #[start] trick doesn’t run anymore.

This runs a test harness nested in another. It’s not pretty, but it works.

[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/168)

### DIFF
--- a/tests/tokenizer.rs
+++ b/tests/tokenizer.rs
@@ -7,7 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#![cfg_attr(feature = "unstable", feature(start, rt, test, plugin))]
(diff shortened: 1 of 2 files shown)


nox:
The something that changed: https://github.com/rust-lang/rust/blob/19a33edaa956a842fed11a1293a5aa7336483f0c/src/libsyntax/test.rs#L205-L207

#[start] and #[main] are now unconditionally removed.

SimonSapin:
I forced-pushed a second too late to add it to the commit message :)

SimonSapin:
@bors-servo r+ df8e749

Don’t make tests fail when they can’t be run.

PR #170, opened 2015-09-03
Update inline tokenize example

johannhof:
`tokenize_to` was at some point changed to move the `sink` value instead of taking a mut ref, but the inline rustdoc example was not updated, causing lots of confusion

[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/170)


SimonSapin:
Thanks!

@bors-servo r+

PR #176, opened 2015-10-11
Fix version support in README

### LINKED ISSUE #171: Update README re: nightly and stable

sfackler:
Closes #171 

[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/176)


SimonSapin:
IIRC all *features* are there, only some optimization (mostly of memory size: `NonZero`, `unsafe_no_drop_flag`, …) require Nightly. (And using a compiler-plugin rather than a pre-generated file, but users don’t care, that only matters when modifying that file.)

So, please change "features" to "optimizations" in the patch, and this looks good.

sfackler:
Ah, thanks for the info, fixed!

PR #183, opened 2015-11-20
Add function to change tokenizer state to Plaintext

### LINKED ISSUE #109: Allow changing tokenizer state to PLAINTEXT
This came from mozilla/servo#4181. The spec for [dealing with text/plain](https://html.spec.whatwg.org/multipage/browsers.html#read-text) has:

> and switch the HTML parser's tokenizer to the PLAINTEXT state.

I'm doing this in doublec/servo@5d4e7efa by using `<plaintext>` as that seems to be the only way of changing to Plaintext.

Can an API be added to allow switching the state for this use case?

KiChjang:
Fixes #109, I'm really not sure whether I'm doing the right thing.

r? @kmcallister 

[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/183)

### DIFF
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -221,6 +221,10 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         self.run();
     }
 
+    pub fn set_plaintext_state(&mut self) {
+        self.state = states::Plaintext;
+    }
+
     fn process_token(&mut self, token: Token) {
         if self.opts.profile {
             let (_, dt) = time!(self.sink.process_token(token));


jdm:
Probably r? @nox instead.

SimonSapin:
Thanks! I published v0.2.10 with this. Do you want to fix https://github.com/servo/servo/commit/82c52a7a1c2d1fcdad797fc73564d6f624650473#diff-14d82a613adf6dbeec0ec4415a1d323aR195 as well?

KiChjang:
That's exactly why I did this. I am working on it as we speak.

SimonSapin:
:+1:

Issue #119, opened 2015-04-04
Parse HTML integration points in MathML


kmcallister:
This is annoying because you have to know [the original attributes that a node was created with](https://html.spec.whatwg.org/multipage/syntax.html#html-integration-point).

nox:
Related tests:

* tb: tests20.dat-34
* tb: tests20.dat-35
* tb: tests20.dat-36
* tb: tests20.dat-37

GuillaumeGomez:
I'll work on it (but will need help).

nox:
This was done by @GuillaumeGomez.

PR #186, opened 2016-01-04
Add is_mathml_annotation_xml_integration_point method

GuillaumeGomez:
r? @nox

#119 

[Review on Reviewable](https://reviewable.io/reviews/servo/html5ever/186)

### DIFF
--- a/src/rcdom.rs
+++ b/src/rcdom.rs
@@ -12,6 +12,7 @@
 //! This is sufficient as a static parse tree, but don't build a
 //! web browser using it. :)
 
+use std::ascii::AsciiExt;
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::default::Default;
@@ -32,7 +33,7 @@ use serialize::TraversalScope;
 use serialize::TraversalScope::{IncludeNode, ChildrenOnly};
 use driver::ParseResult;
 
-pub use self::ElementEnum::{Normal, Script, Template};
+pub use self::ElementEnum::{AnnotationXml, Normal, Script, Template};
 pub use self::NodeEnum::{Document, Doctype, Text, Comment, Element};
 
 /// The different kinds of elements in the DOM.
@@ -45,6 +46,11 @@ pub enum ElementEnum {
     /// A template element and its template contents.
     /// https://html.spec.whatwg.org/multipage/#template-contents
     Template(Handle),
+    /// An annotation-xml element in the MathML namespace whose start tag token had an attribute
+    /// with the name "encoding" whose value was an ASCII case-insensitive match for the string
+    /// "text/html" or "application/xhtml+xml"
+    /// https://html.spec.whatwg.org/multipage/embedded-content.html#math:annotation-xml
+    AnnotationXml(bool),
 }
 
 /// The different kinds of nodes in the DOM.
@@ -198,6 +204,14 @@ impl TreeSink for RcDom {
         let info = match name {
             qualname!(html, "script") => Script(false),
             qualname!(html, "template") => Template(new_node(Document)),
+            qualname!(mathml, "annotation-xml") => {
+                AnnotationXml(attrs.iter().find(|attr| attr.name == qualname!("", "encoding"))
+                                   .map_or(false,
+                                           |attr| attr.value
+                                                      .eq_ignore_ascii_case("text/html") ||
+                                                  attr.value
+                                                      .eq_ignore_ascii_case("application/xhtml+xml")))
+            },
             _ => Normal,
         };
         new_node(Element(name, info, attrs))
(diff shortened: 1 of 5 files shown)


src/rcdom.rs
```diff
@@ -195,6 +196,14 @@ impl TreeSink for RcDom {
     }
 
     fn create_element(&mut self, name: QualName, attrs: Vec<Attribute>) -> Handle {
+        for it in 0..attrs.len() {
+            match attrs[it].name {
+                qualname!(html, "encoding") => {
+                    return new_node(Element(name, AnnotationXml(attrs[it].value.to_lowercase() == "text/html"), attrs));
```
SimonSapin: Use `.eq_ignore_ascii_case("text/html")` to save a memory allocation.

src/rcdom.rs
```diff
@@ -45,6 +45,7 @@ pub enum ElementEnum {
     /// A template element and its template contents.
     /// https://html.spec.whatwg.org/multipage/#template-contents
     Template(Handle),
+    AnnotationXml(bool),
```
SimonSapin: What does it mean for an element to "be" this variant? Is it exclusive with other variants? What does the boolean mean? Consider using a struct-like variant in order to name the boolean. Please add a doc-comment with a link to the relevant section of spec.

src/tree_builder/interface.rs
```diff
@@ -133,6 +133,10 @@ pub trait TreeSink {
     fn complete_script(&mut self, _node: Self::Handle) -> NextParserState {
         NextParserState::Continue
     }
+
+    // Returns true if the adjusted current node is an HTML integration point
+    // and the token is a start tag
+    fn is_mathml_annotation_xml_integration_point(&self) -> bool;
```
nox: This should take a `Handle` and that's the thing on which you check whether it is an integration point or not.

src/rcdom.rs
```diff
@@ -303,6 +312,13 @@ impl TreeSink for RcDom {
             panic!("not a script element!");
         }
     }
+
+    fn is_mathml_annotation_xml_integration_point(&self) -> bool {
+        match (**self.document).borrow().node {
```
nox: The thing you need to check is the given `Handle` that needs to be passed to the method.

src/rcdom.rs
GuillaumeGomez: I didn't add comments yet, it was just supposed to be a first shot to get help. on some details before finishing it. I should have precised, my bad.

GuillaumeGomez:
Ok, I updated the current code @nox.

src/rcdom.rs
```diff
@@ -303,6 +313,13 @@ impl TreeSink for RcDom {
             panic!("not a script element!");
         }
     }
+
+    fn is_mathml_annotation_xml_integration_point(&self, handle: &Self::Handle) -> bool {
+        match (**handle).borrow().node {
+            Element(_, AnnotationXml(ret), _) => ret,
+            _ => false,
```
nox: This should be unreachable.

src/rcdom.rs
```diff
@@ -195,6 +197,14 @@ impl TreeSink for RcDom {
...
 
     fn create_element(&mut self, name: QualName, attrs: Vec<Attribute>) -> Handle {
+        for it in 0..attrs.len() {
+            match attrs[it].name {
+                qualname!(html, "encoding") => {
+                    return new_node(Element(name, AnnotationXml(attrs[it].value.eq_ignore_ascii_case("text/html")), attrs));
+                }
+                _ => continue,
+            }
+        }
```
nox: This should be in the match expression defining `info` below, for `qualname!(mathml, "annotation-xml")` elements.

GuillaumeGomez:
Anything else to add/update?

src/tree_builder/actions.rs
```diff
@@ -901,11 +901,15 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
 
         if let qualname!(mathml, "annotation-xml") = name {
             if let TagToken(Tag { kind: StartTag, name: atom!("svg"), .. }) = *token {
```
nox: This condition should go away. It has nothing to do here.

src/tree_builder/interface.rs
```diff
@@ -133,6 +133,12 @@ pub trait TreeSink {
     fn complete_script(&mut self, _node: Self::Handle) -> NextParserState {
         NextParserState::Continue
     }
+
+    // Returns true if the adjusted current node is an HTML integration point
+    // and the token is a start tag
+    fn is_mathml_annotation_xml_integration_point(&self, handle: &Self::Handle) -> bool {
```
nox: This should take `Self::Handle`, and you should clone handles when passing them.

src/tree_builder/actions.rs
nox: I was wrong about this.

Issue #203, opened 2016-03-24
C API references non-existing "string_cache.h"


RReverser:
`html5ever.h` [references](https://github.com/servo/html5ever/blob/master/capi/include/html5ever.h#L15) `string_cache.h` although that definition file can't be found neither in this repo nor in https://github.com/servo/string-cache.

Also, https://github.com/servo/html5ever/blob/master/examples/capi/tokenize.c example uses `struct h5e_buf` in place of `scache_atom`s in `html5ever.h` and undefined function `h5e_buf_from_cstr`, so I suspect header is just outdated and needs to define new structure instead of including `string_cache.h`.

**UPD**: Actually, if I understand https://github.com/servo/html5ever/blob/master/capi/src/tokenizer.rs#L121 correctly, those functions accept just Tendrils (instead of `scache_atom` / `h5e_buf`), so it should be possible to change definitions like following:

```c
void (*do_start_tag)(void *user, tendril name, int self_closing, size_t num_attrs);
void (*do_tag_attr)(void *user, tendril name, tendril value);
void (*do_end_tag)(void *user, tendril name);
```

and then code like this would work:

```c
tendril buf = TENDRIL_INIT;
tendril_push_buffer(&buf, html, size);
h5e_tokenizer_feed(tok, buf);
```

(although I might be wrong)

Ygg01:
kmcallister (kmc) worked on it, before they left. It was here at  https://github.com/servo/string-cache/pull/84 but was never merged with string-cache master. It has bitroted somewhat (May 2015). 

I'm willing to work on it, tomorrow, since it's one of the rare days I have time to commit to it.

asajeffrey:
Note #204

RReverser:
@Ygg01 Sure, no rush. I'm actually already trying to fix it and come up with PR (no promises yet though).

RReverser:
@asajeffrey I'm confused as to why it's being removed :/ Hoped for the opposite (extending).

Ygg01:
@RReverser If push comes to shove, just copy the capi into a separate crate and publish it independently. You'll probably need to something similar to string-cache.

RReverser:
@Ygg01 Actually that PR to string-cache looks good to me. Need to try as to whether it still works.

SimonSapin:
I don’t think there’s anything worth salvaging in `html5ever.h`. Closing in favor of #204, please keep the discussion there.

RReverser:
[...]

PR #204, opened 2016-03-24
Remove the C API

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/204)


SimonSapin:
```
+cargo test --manifest-path capi/Cargo.toml
error: manifest path `capi/Cargo.toml` does not exist
```

Ygg01:
Why is C API being removed?

nox:
Because, to quote @SimonSapin:

> it’s been broken for a long time, and it only has tokenization

And also:

> 14:55 <•SimonSapin> ManishCloud jdm nox: I have some experiments at https://github.com/SimonSapin/html5ever-python
> 14:55 <•SimonSapin> python parts aside this includes a C-compatible ABI

RReverser:
> it’s been broken for a long time, and it only has tokenization

Well, then fixing and extending is perhaps a better option that removing?

nox:
It's been broken for a long time and nobody cared.

RReverser:
@nox Well, I do care because want to use html5ever from our codebase. As said in #203, actually trying to fix at the moment.

SimonSapin:
@RReverser you likely want to use the parser (tree builder) and don’t care about the tokenizer in isolation. At this point what would be needed is more of a complete rewrite than fixing up existing stuff.

If you want to invest in this I’d recommend starting with a copy of https://github.com/SimonSapin/html5ever-python/blob/master/rust-glue/lib.rs . It doesn’t have to be in the servo/html5ever repository.

As you can see, in that C API strings are only exposed as (pointer, length) pairs. The details of string-cache and tendril are kept within Rust-land. This is a choice I made to manage the complexity. Depending on the application a different choice might make sense.

RReverser:
@SimonSapin Actually I'm interested exactly in tokenizer at the moment (as mentioned in previous issues).

PR #205, opened 2016-04-06
Fixed compilation errors in macros. Whitespace handling in Tree Builder.

martinrlilja:
Hi!

I’m working on a project which uses html5ever and I noticed that the tree builder handles newlines and whitespace outside of the `body` and `head` tags in a slightly odd way.

Take a simple document like this:

```html
<!DOCTYPE html>
<html>
    <head></head>
    <body></body>
</html>
```

Currently html5ever ignores the newline and indentation of head. For example, the print-rcdom example yields this on the above document:

```
#Document
    <!DOCTYPE html "" "">
    <html>
        <head>
        #text: \n    
        <body>
            #text: \n\n
```

So I changed the `BeforeHead` rule in `tree_builder/rules.rs` to treat text in the same way `AfterHead` does.
(As a side note, doing so yielded some errors in the macros crate which I fixed. Those fixes are also part of this pr.)

```
#Document
    <!DOCTYPE html "" "">
    <html>
        #text: \n    
        <head>
        #text: \n    
        <body>
            #text: \n\n
```

So there is now a newline and indentation between the `html` and `head` tags. Perfect! But there are still two newlines inside the empty `body` tag. (Both of them exists after the end tag of body.) I’m guessing they come from the `AfterBody` rule. Which, I guess, is there in case there is text after the body. But it seems odd that it affects a valid document like this.

So without knowing the spec, maybe only text up until the last non-whitespace character should be inserted into the body and the rest should be where it originally occurred in the input stream.

The output should then look like this:

```
#Document
    <!DOCTYPE html "" "">
    <html>
        #text: \n    
        <head>
        #text: \n    
        <body>
        #text: \n
    #text: \n
```

Does this seem reasonable?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/205)

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -18,10 +18,9 @@ extern crate mac;
 // See https://github.com/rust-lang/rust/pull/23857
 macro_rules! panictry {
     ($e:expr) => ({
-        use syntax::diagnostic::FatalError;
         match $e {
             Ok(e) => e,
-            Err(FatalError) => panic!(FatalError)
+            Err(err) => panic!("{:?}", err),
         }
     })
 }
--- a/macros/src/match_token.rs
+++ b/macros/src/match_token.rs
@@ -100,12 +100,12 @@ matching, by enforcing the following restrictions on its input:
 use std::collections::{HashSet, HashMap};
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 
-use syntax::diagnostic::FatalError;
+use syntax::errors::FatalError;
 use syntax::ptr::P;
 use syntax::codemap::{Span, Spanned, spanned};
 use syntax::ast;
 use syntax::parse::parser::{Parser, Restrictions};
-use syntax::parse::{token, parser, classify};
+use syntax::parse::{token, parser, classify, PResult};
 use syntax::parse;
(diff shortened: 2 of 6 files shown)


nox:
If you look at the spec, you'll see it tells to do exactly what we do:

https://html.spec.whatwg.org/multipage/syntax.html#the-before-head-insertion-mode

> * A character token that is one of U+0009 CHARACTER TABULATION, U+000A LINE FEED (LF), U+000C FORM FEED (FF), U+000D CARRIAGE RETURN (CR), or U+0020 SPACE

>   * Ignore the token.

nox:
What was the first commit for, btw?

martinrlilja:
Oh, thanks, should’ve ran the tests :/

When I changed `tree_builder/rules.rs` it wanted me to compile with `cargo build --features codegen` which gave a bunch of errors because apparently `syntax::Parser` functions no longer return a `FatalError`.

jdm:
Shoot, I'm really sorry we dropped the ball on reviewing these changes. This would have avoided #216 (or at least alerted us sooner); as it is, we're replacing the macros in #217, so I'm going to go ahead and close this. Sorry for the frustrating experience!

PR #206, opened 2016-04-14
Remove blank line after serializing doctype

jaysonsantos:
As said here [1], the line break after the doctype is undesirable.

[1] https://github.com/servo/servo/pull/10604#issuecomment-209942021

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/206)


jdm:
This matches my reading of https://html.spec.whatwg.org/multipage/syntax.html#serialising-html-fragments, which is good. Oddly, I don't see any historical reason for including the newline when looking at older revisions of the specification, even around the time that this was originally implemented.

jaysonsantos:
Thanks!

PR #209, opened 2016-04-26
Make get_template_contents take a mutable self argument.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/209)


SimonSapin:
Seems reasonable, but out of curiosity: why?

KiChjang:
```
<jdm> KiChjang: it would make a bunch of code in https://github.com/servo/servo/compare/master...jdm:asyncparse much nicer
```

PR #212, opened 2016-07-11
html2html is a very helpful example for parsing and then serializing html

iopq:
I had some troubles with doing this and I didn't realize that a minimal example already exists. The first example does too much stuff so it's not very easy to get started.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/212)


jdm:
@bors-servo: r+
Thanks for making this change!

PR #213, opened 2016-08-20
Use std::time instead of deprecated crate time

spk:
Crate time is deprecated; switch to `std::time`
Cheers

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/213)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ heap_size = ["heapsize", "heapsize_plugin"]
 codegen = ["html5ever_macros"]
 
 [dependencies]
-time = "0"
 log = "0"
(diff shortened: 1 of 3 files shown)


jdm:
This change reduces the accuracy of the timer from nanoseconds to seconds, which is much less useful. Can we improve on that?

SimonSapin:
I’m making a PR to Rust’s standard library to add a `Duration` method doing this, but in the meantime please use `d.as_secs() * 1_000_000_000 + u64::from(d.subsec_nanos())`

SimonSapin:
https://github.com/rust-lang/rust/pull/35868

PR #217, opened 2016-10-03
Use syn and quote crates for the `match_token!` macro…

### LINKED ISSUE #216: Macros to generate tree building rules don't work in current Rust
Apparently the code was generated long enough ago that rules around writing macros have changed.
```
<pcwalton> ugh, and I can’t rebuild html5ever macros because it’s incompatible with my rustc
<pcwalton> we can’t make changes to the tree building rules because nobody updated the macros to work with current rust
 <jdm> pcwalton: sounds like building those macros should be part of CI
```

SimonSapin:
… instead of the parser and quasi-quoting from Rust’s (unstable) libsyntax.

This has significantly worse diagnostics when encountering unexpected syntax (e.g. no indication of which line has the offending code) but this removes all usage of unstable compiler internals that constantly need to be fixed when updating the compiler.

Fixes #216.

r? @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/217)


Cargo.toml
```diff
@@ -33,7 +33,7 @@ harness = false
 [features]
 unstable = ["tendril/unstable", "string_cache/unstable"]
 heap_size = ["heapsize", "heapsize_plugin"]
-codegen = ["html5ever_macros"]
+codegen = []
```
SimonSapin: This feature is now unused, but removing it would be a breaking change. So let’s keep it until we have some other reason to make a breaking change.

src/tree_builder/rules.rs
```diff
@@ -7,7 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-//! The tree builder rules, as a single, enormous nested match expression.
```
SimonSapin: `include!` doesn’t like top-level attributes.

src/tree_builder/rules.rs
```diff
@@ -770,7 +770,7 @@ impl<Handle, Sink> TreeBuilderStep
                     Done
                 }
 
-                // FIXME: This should be unreachable, but match_token! requires a
+                // FIXME: This should be unreachable, but match_token requires a
```
SimonSapin: Yes, the parser is that stupid. It would find `match_token!` in a comment and try to parse it as a macro invocation.

macros/match_token.rs
```diff
@@ -0,0 +1,505 @@
...
+                b'[' => i = self.find_top_level(i, b']'),
+                b'(' => i = self.find_top_level(i, b')'),
+                _ => {}
+            }
+        }
+    }
+}
+
+pub fn expand_match_tokens(from: &Path, to: &Path) {
+//    use std::fmt::Write;
```
emilio: Do you plan to keep these comments for any specific reasons? Or is it fine to remove them?

SimonSapin: They were temporary, for debugging. I forgot to remove them before committing. Removed now.

SimonSapin:
> Are there any tests for this macro?

Not directly, but there are many tree builder tests that exercise the generated code.

Ygg01:
How easy will this be the modify?

nox:
syn 0.9.0 is out with many improvements, could the hand-written parser use it? I'm really wary of merging something that can also end up transforming comments.

Ygg01:
@nox: `syn` is for derive macros, only, as far as I can tell. See discussion here: https://github.com/servo/html5ever/issues/216

nox:
@Ygg01 https://github.com/dtolnay/syn/releases/tag/0.9.0

Ygg01:
@nox: I'm a newbie when it comes to procedural macros, but those notes don't show me anything that would make full procedural macros parsable by `syn`? I hate to be an energy vampire, so could you just explain what you meant?

Did you mean we could parse tags as:

```rust
match mode {
    InitialMode => 
    #[derive(tags)]
    match token {
         CharacterTokens(NotSplit, text) => SplitWhitespace(text),
         ...
         token => {
                    if !self.opts.iframe_srcdoc {
                        self.unexpected(&token);
                        self.set_quirks_mode(Quirks);
                    }
                    Reprocess(BeforeHtml, token)
          }
    }
}
```

SimonSapin:
Alright! I pushed a rewrite that uses the `syn` and `quote` crates, the provide real parser and quasi-quoting.

r? @nox 

I’ll rebase and fix conflicts tomorrow.

src/tree_builder/rules.rs
SimonSapin: Note: this comment referred to my hand-written parser. `syn` doesn’t have this problem.

PR #224, opened 2016-10-23
Properly encode the states for CDATA

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/224)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "html5ever"
-version = "0.6.1"
+version = "0.7.0"
 authors = [ "The html5ever Project Developers" ]
 license = "MIT / Apache-2.0"
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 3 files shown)


SimonSapin:
`tb: tests21.dat-11 (scripting disabled)` is failing on Travis.

Did the spec change? Should html5lib-tests be updated accordingly?

Is this a breaking change because the `State` enum is public? Does it need to be? More generally, could we make the public API smaller so that we can change internals without breaking it?

nox:
It needs to be because the tree builder can change the state of the tokenizer. To any state. That should be improved but this is still a breaking change.

Will look into the failing test, that wasn't supposed to happen and I didn't even see it locally...

nox:
@SimonSapin I fixed the bug.

SimonSapin:
IRC:
> * nox As for whether the spec changed, no idea, but I would rather we follow the states closely, and in the spec the CDATA states don't use look-ahead, so I think we should align ourselves.
> * SimonSapin: so it shouldn’t change the overall behavior of tokenizer + tree builder ?
> * nox: Yep.

@bors-servo r+

PR #227, opened 2016-11-02
fix benchmarking broken by  "Make tokenizer not own the input stream"

jackinloadup:
Was curious about project benchmarks and noticed it was broken. A quick bisect revealed it was recently broken in 7fec1da. Thankfully there was examples of testing that needed to be updated which correlated directly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/227)


nox:
Oops, thanks!

@bors-servo r+

PR #240, opened 2016-12-01
Added current_line variable to Tokenizer and test to check that curre…

karenher:
…nt_line is correctly updated

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/240)

### DIFF
--- a/benches/tokenizer.rs
+++ b/benches/tokenizer.rs
@@ -29,7 +29,7 @@ struct Sink;
 impl TokenSink for Sink {
     type Handle = ();
 
-    fn process_token(&mut self, token: Token) -> TokenSinkResult<()> {
+    fn process_token(&mut self, token: Token, line_number: u64) -> TokenSinkResult<()> {
         // Don't use the token, but make sure we don't get
         // optimized out entirely.
         black_box(token);
--- a/benches/tokenizer.rs
+++ b/benches/tokenizer.rs
@@ -29,7 +29,7 @@ struct Sink;
 impl TokenSink for Sink {
     type Handle = ();
 
-    fn process_token(&mut self, token: Token, line_number: u64) -> TokenSinkResult<()> {
+    fn process_token(&mut self, token: Token, _line_number: u64) -> TokenSinkResult<()> {
         // Don't use the token, but make sure we don't get
         // optimized out entirely.
         black_box(token);
--- a/src/rcdom.rs
+++ b/src/rcdom.rs
@@ -161,6 +161,9 @@ pub struct RcDom {
 
     /// The document's quirks mode.
     pub quirks_mode: QuirksMode,
+    
+    /// The current line being parsed
+    pub current_line: u64,
 }
 
 impl TreeSink for RcDom {
@@ -326,6 +329,10 @@ impl TreeSink for RcDom {
             _ => unreachable!(),
         }
     }
+
+    fn set_current_line(&mut self, line_number: u64) {
+        self.current_line = line_number;
+    }
 }
 
 impl Default for RcDom {
@@ -334,6 +341,7 @@ impl Default for RcDom {
             document: new_node(Document),
             errors: vec!(),
             quirks_mode: tree_builder::NoQuirks,
+            current_line: 1,
         }
     }
 }
--- a/src/rcdom.rs
+++ b/src/rcdom.rs
@@ -161,9 +161,6 @@ pub struct RcDom {
 
     /// The document's quirks mode.
     pub quirks_mode: QuirksMode,
-    
-    /// The current line being parsed
-    pub current_line: u64,
 }
 
 impl TreeSink for RcDom {
@@ -330,9 +327,7 @@ impl TreeSink for RcDom {
         }
     }
 
-    fn set_current_line(&mut self, line_number: u64) {
-        self.current_line = line_number;
-    }
+    fn set_current_line(&mut self, _line_number: u64) { }
 }
 
 impl Default for RcDom {
@@ -341,7 +336,6 @@ impl Default for RcDom {
             document: new_node(Document),
             errors: vec!(),
             quirks_mode: tree_builder::NoQuirks,
-            current_line: 1,
         }
     }
 }
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -698,7 +698,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         match self.state {
             //§ data-state
             states::Data => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0' '&' '<')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '&' '<' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\0'),
                     FromSet('&')  => go!(self: consume_char_ref),
                     FromSet('<')  => go!(self: to TagOpen),
@@ -709,7 +709,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ rcdata-state
             states::RawData(Rcdata) => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0' '&' '<')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '&' '<' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\u{fffd}'),
                     FromSet('&') => go!(self: consume_char_ref),
                     FromSet('<') => go!(self: to RawLessThanSign Rcdata),
@@ -720,7 +720,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ rawtext-state
             states::RawData(Rawtext) => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0' '<')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '<' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\u{fffd}'),
                     FromSet('<') => go!(self: to RawLessThanSign Rawtext),
                     FromSet(c) => go!(self: emit c),
@@ -730,7 +730,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ script-data-state
             states::RawData(ScriptData) => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0' '<')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '<' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\u{fffd}'),
                     FromSet('<') => go!(self: to RawLessThanSign ScriptData),
                     FromSet(c) => go!(self: emit c),
@@ -740,7 +740,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ script-data-escaped-state
             states::RawData(ScriptDataEscaped(Escaped)) => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0' '-' '<')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '-' '<' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\u{fffd}'),
                     FromSet('-') => go!(self: emit '-'; to ScriptDataEscapedDash Escaped),
                     FromSet('<') => go!(self: to RawLessThanSign ScriptDataEscaped Escaped),
@@ -751,7 +751,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ script-data-double-escaped-state
             states::RawData(ScriptDataEscaped(DoubleEscaped)) => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0' '-' '<')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '-' '<' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\u{fffd}'),
                     FromSet('-') => go!(self: emit '-'; to ScriptDataEscapedDash DoubleEscaped),
                     FromSet('<') => go!(self: emit '<'; to RawLessThanSign ScriptDataEscaped DoubleEscaped),
@@ -762,7 +762,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ plaintext-state
             states::Plaintext => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '\0')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '\0' '\n')) {
                     FromSet('\0') => go!(self: error; emit '\u{fffd}'),
                     FromSet(c)    => go!(self: emit c),
                     NotFromSet(b) => self.emit_chars(b),
@@ -982,7 +982,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             //§ attribute-value-(double-quoted)-state
             states::AttributeValue(DoubleQuoted) => loop {
-                match pop_except_from!(self, input, small_char_set!('\r' '"' '&' '\0')) {
+                match pop_except_from!(self, input, small_char_set!('\r' '"' '&' '\0' '\n')) {
                     FromSet('"')  => go!(self: to AfterAttributeValueQuoted),
                     FromSet('&')  => go!(self: consume_char_ref '"'),
                     FromSet('\0') => go!(self: error; push_value '\u{fffd}'),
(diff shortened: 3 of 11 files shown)


src/tokenizer/mod.rs
```diff
@@ -200,6 +203,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
             temp_buf: StrTendril::new(),
             state_profile: BTreeMap::new(),
             time_in_sink: 0,
+	    current_line: 1,
```
jdm: nit: replace the tabs in this line with spaces.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
     }
 }
 
+
+
+
+
```
jdm: nit: remove these new blank lines.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
 mod test {
     use super::option_push; // private items
     use tendril::{StrTendril, SliceExt};
 
+    use std::io::{self};
+    use super::{TokenSink, Tokenizer, TokenizerOpts, TokenSinkResult};
+
+    pub use super::interface::{Token, DoctypeToken, TagToken, CommentToken};
+    pub use super::interface::{CharacterTokens, NullCharacterToken, EOFToken, ParseError};
+    pub use super::interface::{Doctype, Attribute, TagKind, StartTag, EndTag, Tag};
```
jdm: The `pub` in these lines isn't necessary.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+    use std::mem::replace;
+    use std::borrow::Cow::{self, Borrowed};
+
+    // TokenMatch implements the TokenSink trait. It is used for testing to see 
+    // if current_line is being updated when process_token is called. The lines
+    // vector is a collection of the line numbers that each token is on.
+    struct TokenMatch {
+        tokens: Vec<Token>,
+        current_str: StrTendril,
+        exact_errors: bool,
```
jdm: We don't need the `tokens` or `exact_errors` members for this test.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+    pub use super::interface::{Doctype, Attribute, TagKind, StartTag, EndTag, Tag};
+
+    use super::buffer_queue::{BufferQueue};
+    use std::mem::replace;
+    use std::borrow::Cow::{self, Borrowed};
+
+    // TokenMatch implements the TokenSink trait. It is used for testing to see 
+    // if current_line is being updated when process_token is called. The lines
+    // vector is a collection of the line numbers that each token is on.
+    struct TokenMatch {
```
jdm: Let's call this `LinesMatch` instead, since that's what it's testing.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+            }
+        }
+
+    }
+
+    impl TokenSink for TokenMatch {
+        
+        type Handle = ();
+
+        fn process_token(&mut self, token: Token, line_number: u64) -> TokenSinkResult<Self::Handle>{
```
jdm: nit: add a space before {

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+    use std::borrow::Cow::{self, Borrowed};
+
+    // TokenMatch implements the TokenSink trait. It is used for testing to see 
+    // if current_line is being updated when process_token is called. The lines
+    // vector is a collection of the line numbers that each token is on.
+    struct TokenMatch {
+        tokens: Vec<Token>,
+        current_str: StrTendril,
+        exact_errors: bool,
+        lines: Vec<u64>,
```
jdm: Let's make this `Vec<(Token, u64)>`, so we can verify that the tokens produced are what we expect.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+            match token {
+                CharacterTokens(b) => {
+                    self.current_str.push_slice(&b);
+                }
+
+                NullCharacterToken => {
+                    self.current_str.push_char('\0');
+                }
+
+                ParseError(_) => if self.exact_errors {
```
jdm: We can use `ParseError(_) => panic!("unexpected parse error")` for this test.

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+
+    }
+
+    impl TokenSink for TokenMatch {
+        
+        type Handle = ();
+
+        fn process_token(&mut self, token: Token, line_number: u64) -> TokenSinkResult<Self::Handle>{
+            
+            self.lines.push(line_number);
```
jdm: Let's move this into the `push` method (passing line_number as an argument).

src/tokenizer/mod.rs
```diff
@@ -1397,12 +1402,118 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+
+                _ => self.push(token),
+            }
+            TokenSinkResult::Continue
+        }
+    }
+
+    // Take in tokens, process them, and return vector with line 
+    // numbers that each token is on
+    fn tokenize(input: Vec<StrTendril>, opts: TokenizerOpts) -> Vec<u64> {
```
jdm: Let's return `Vec<(Token, u64)>`.

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1534,20 @@ mod test {
         option_push(&mut s, 'x');
         assert_eq!(s, Some("yx".to_tendril()));
     }
+
+    #[test]
+    fn check_lines() {
+
```
jdm: nit: remove this empty line.

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1534,20 @@ mod test {
...
+        let opts = TokenizerOpts{exact_errors: false, discard_bom: true, profile: false,
```
jdm: nit: space before and after {

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1534,20 @@ mod test {
...
+            initial_state: None, last_start_tag_name: None,};
```
jdm: nit: move `};` to the next line, like:
```rust
let opts = TokenizerOpts { ...,
    ...,
};
```

src/tokenizer/mod.rs
```diff
@@ -270,6 +274,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         if c == '\r' {
             self.ignore_lf = true;
             c = '\n';
+	    self.current_line += 1;
```
jdm: Let's make a separate block after this instead:
```rust
if c == '\r' {
    self.ignore_lf = true;
    c = '\n';
}

if c == '\n' {
    self.current_line += 1;
}
```
This means that input that only contains \n tokens will correctly count lines, and code that contains \r\n tokens (which is what the ignore_lf checks are for) will also count lines correctly.

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1534,20 @@ mod test {
...
     }
+
+    #[test]
+    fn check_lines() {
+
+        let opts = TokenizerOpts{exact_errors: false, discard_bom: true, profile: false,
+            initial_state: None, last_start_tag_name: None,};
+
+        let vector = vec![StrTendril::from("<a>\r"), StrTendril::from("<b>\r"),
+            StrTendril::from("</b>\r"), StrTendril::from("</a>\r")];
```
jdm: We should run this test twice - once using `\n` separators, and once using `\r\n`.

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1534,20 @@ mod test {
...
+        let expected = vec![1, 2, 2, 3, 3, 4, 4, 5, 5];
```
jdm: We'll need to update this to contain pairs of tokens and the corresponding line number.

benches/tokenizer.rs
```diff
@@ -29,7 +29,7 @@ struct Sink;
 impl TokenSink for Sink {
     type Handle = ();
 
-    fn process_token(&mut self, token: Token) -> TokenSinkResult<()> {
+    fn process_token(&mut self, token: Token, line_number: u64) -> TokenSinkResult<()> {
```
jdm: Let's use `_line_number` here and elsewhere to avoid warnings about unused variables in code that never makes use of it.

src/rcdom.rs
```diff
@@ -161,6 +161,9 @@ pub struct RcDom {
 
     /// The document's quirks mode.
     pub quirks_mode: QuirksMode,
+    
+    /// The current line being parsed
+    pub current_line: u64,
```
jdm: Let's remove this, since no RcDom code actually uses it.

src/rcdom.rs
```diff
@@ -326,6 +329,10 @@ impl TreeSink for RcDom {
             _ => unreachable!(),
         }
     }
+
+    fn set_current_line(&mut self, line_number: u64) {
+        self.current_line = line_number;
```
jdm: This can be an empty method instead (and use `_line_number` to avoid warnings).

src/tokenizer/mod.rs
```diff
@@ -1403,6 +1411,118 @@ mod test {
...
+            LinesMatch {
+                tokens: vec!(),
+                current_str: StrTendril::new(),
+                lines: vec!(),
+            }
+        }
+
+        fn push(&mut self, token: Token, line_number: u64) {
+            self.finish_str();
+            // self.tokens.push(token);
```
jdm: Let's remove this.

src/tokenizer/mod.rs
```diff
@@ -1403,6 +1411,118 @@ mod test {
...
+    use std::io::{self};
+    use super::{TokenSink, Tokenizer, TokenizerOpts, TokenSinkResult};
+
+    use super::interface::{Token, DoctypeToken, TagToken, CommentToken};
+    use super::interface::{CharacterTokens, NullCharacterToken, EOFToken, ParseError};
+    use super::interface::{Doctype, Attribute, TagKind, StartTag, EndTag, Tag};
+
+    use super::buffer_queue::{BufferQueue};
+    use std::mem::replace;
+    use std::borrow::Cow::{self, Borrowed};
```
jdm: We don't use Cow anywhere, so this import can be removed.

src/tokenizer/mod.rs
```diff
@@ -1403,6 +1411,118 @@ mod test {
     use super::option_push; // private items
     use tendril::{StrTendril, SliceExt};
 
+    use std::io::{self};
```
jdm: We don't use io anywhere in this test, so this can be removed.

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1543,34 @@ mod test {
         option_push(&mut s, 'x');
         assert_eq!(s, Some("yx".to_tendril()));
     }
+
+    #[test]
+    fn check_lines() {
+        let opts = TokenizerOpts { exact_errors: false, discard_bom: true, profile: false,
+            initial_state: None, last_start_tag_name: None,
+        };
```
jdm: nit: let's do:
```rust
let opts = TokenizerOpts {
    exact_errors: false,
    discard_bom: true,
    //etc.
};
```

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1543,34 @@ mod test {
...
+        let vector = vec![StrTendril::from("<a>\r"), StrTendril::from("<b>\r"),
+            StrTendril::from("</b>\r"), StrTendril::from("</a>\r")];
```
jdm: We'll want to use `\n` here instead; all real content will either use `\n` or `\r\n`.

src/tokenizer/mod.rs
```diff
@@ -1423,4 +1543,34 @@ mod test {
...
+            (create_tag(StrTendril::from("a"), EndTag), 4)];
+        let results = tokenize(vector, opts);
+        assert_eq!(results, expected);
+    }
+
+    #[test]
+    fn check_lines_with_new_line() {
+        let opts = TokenizerOpts { exact_errors: false, discard_bom: true, profile: false,
+            initial_state: None, last_start_tag_name: None,
+        };
```
jdm: Same style comment as previous.

src/tree_builder/mod.rs
```diff
@@ -33,6 +33,13 @@ use std::mem::replace;
 use std::borrow::Cow::Borrowed;
 use std::collections::VecDeque;
 
+pub use rcdom::ElementEnum::{AnnotationXml, Normal, Template};
+pub use rcdom::NodeEnum::{Document, Comment};
+use std::cell::RefCell;
+use std::rc::{Rc, Weak};
+
+use rcdom::{Node, Handle, RcDom, NodeEnum, ElementEnum};
```
jdm: It's not clear to me why these new imports are required here.

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+            self.rcdom.mark_script_already_started(target);
+        }
+
+        fn is_mathml_annotation_xml_integration_point(&self, handle: Self::Handle) -> bool {
+            self.rcdom.is_mathml_annotation_xml_integration_point(handle)
+        }
+
+        fn set_current_line(&mut self, line_number: u64) {
+            self.set_current_line(line_number);
+            self.rcdom.set_current_line(line_number);
```
jdm: No need to call through to rcdom here.

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+        fn mark_script_already_started(&mut self, target: Handle) {
+            self.rcdom.mark_script_already_started(target);
+        }
+
+        fn is_mathml_annotation_xml_integration_point(&self, handle: Self::Handle) -> bool {
+            self.rcdom.is_mathml_annotation_xml_integration_point(handle)
+        }
+
+        fn set_current_line(&mut self, line_number: u64) {
+            self.set_current_line(line_number);
```
jdm: This will trigger endless recursion, since it's calling itself.

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+
+    use driver::*;
+    use super::{TreeBuilderOpts, TreeBuilder};
+    use tokenizer::Attribute;
+    use rcdom::{Node, Handle, RcDom, NodeEnum, ElementEnum};
+
+    pub struct LineCountingDOM {
+        pub line_vec: Vec<(QualName, u64)>,
+        pub rcdom: RcDom,
+    }
```
jdm: Let's store the current line in this structure, rather than RcDom.

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+            self.rcdom.is_mathml_annotation_xml_integration_point(handle)
+        }
+
+        fn set_current_line(&mut self, line_number: u64) {
+            self.set_current_line(line_number);
+            self.rcdom.set_current_line(line_number);
+        }
+    }
+
+    impl Default for LineCountingDOM {
```
jdm: We don't need this implementation; we can write out the initializer by hand instead of using LineCountingDOM::default().

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+    #[test]
+    fn check_four_lines() {
+        // Input
+        let sink = LineCountingDOM::default();
+        let opts = ParseOpts::default();
+        let mut resultTok = parse_document(sink, opts);
+        resultTok.process(StrTendril::from("<a>"));
+        resultTok.process(StrTendril::from("</a>"));
+        resultTok.process(StrTendril::from("<b>"));
+        resultTok.process(StrTendril::from("</b>"));
```
jdm: We're going to want some newlines in this test, or it's not actually doing much good ;)

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+        // Actual Output
+        let actual = resultTok.finish();
+        // Expected Output
+        let expected = vec![(qualname!(html, "html"), 1),
+                            (qualname!(html, "head"), 1),
+                            (qualname!(html, "body"), 1),
+                            (qualname!(html, "a"), 1),
+                            (qualname!(html, "b"), 1)];
```
jdm: Our expected output will want to ensure that at least some elements are associated with different lines.

src/tree_builder/mod.rs
```diff
@@ -435,3 +450,157 @@ impl<Handle, Sink> TokenSink
...
+        let result = actual.line_vec.clone();
```
jdm: Why is this necessary?

PR #247, opened 2017-01-05
Improve handling of stack of open elements

nox:
Needed for "the end" and proper parsing of style elements.


Ms2ger:
r+ if you use `rposition`

PR #248, opened 2017-01-05
Properly call TokenSink::end from Tokenizer::end

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/248)


Ms2ger:
I guess.

@bors-servo r+

PR #249, opened 2017-01-23
Implement support for form owner

canaltinova:
This is rebased version of #137. Fixed merge conflicts, updated to current codebase and addressed some comments. But there are still some todo's needs to be addressed. Servo side of this changes is currently WIP. Opening this for early feedbacks.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/249)

### DIFF
--- a/src/tree_builder/actions.rs
+++ b/src/tree_builder/actions.rs
@@ -210,40 +210,8 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
 
     // Insert at the "appropriate place for inserting a node".
     fn insert_appropriately(&mut self, child: NodeOrText<Handle>, override_target: Option<Handle>) {
-        declare_tag_set!(foster_target = "table" "tbody" "tfoot" "thead" "tr");
-        let target = override_target.unwrap_or_else(|| self.current_node());
-        if !(self.foster_parenting && self.elem_in(target.clone(), foster_target)) {
-            if self.html_elem_named(target.clone(), local_name!("template")) {
-                // No foster parenting (inside template).
-                let contents = self.sink.get_template_contents(target);
-                self.sink.append(contents, child);
-            } else {
-                // No foster parenting (the common case).
-                self.sink.append(target, child);
-            }
-            return;
-        }
-
-        // Foster parenting
-        let mut iter = self.open_elems.iter().rev().peekable();
-        while let Some(elem) = iter.next() {
-            if self.html_elem_named(elem.clone(), local_name!("template")) {
-                let contents = self.sink.get_template_contents(elem.clone());
-                self.sink.append(contents, child);
-                return;
-            } else if self.html_elem_named(elem.clone(), local_name!("table")) {
-                // Try inserting "inside last table's parent node, immediately before last table"
-                if let Err(child) = self.sink.append_before_sibling(elem.clone(), child) {
-                    // If last_table has no parent, we regain ownership of the child.
-                    // Insert "inside previous element, after its last child (if any)"
-                    let previous_element = (*iter.peek().unwrap()).clone();
-                    self.sink.append(previous_element, child);
-                }
-                return;
-            }
-        }
-        let html_elem = self.html_elem();
-        self.sink.append(html_elem, child);
+        let insertion_point = self.appropriate_place_for_insertion(override_target);
+        self.insert_at(insertion_point, child);
     }
 
     fn adoption_agency(&mut self, subject: LocalName) {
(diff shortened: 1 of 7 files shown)


src/tree_builder/interface.rs
```diff
@@ -72,6 +72,10 @@ pub trait TreeSink {
     /// Do two handles refer to the same node?
     fn same_node(&self, x: Self::Handle, y: Self::Handle) -> bool;
 
+    /// Are two handles present in the same tree
+    /// https://html.spec.whatwg.org/multipage/infrastructure.html#home-subtree
```
jdm: This link isn't valid any more. The text in step 12 of https://html.spec.whatwg.org/multipage/syntax.html#create-an-element-for-the-token just says "same tree" with a link to https://dom.spec.whatwg.org/#concept-tree; we can probably just name this `same_tree` in that case.

src/tree_builder/actions.rs
```diff
@@ -775,10 +743,39 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
         // FIXME: application cache selection algorithm
     }
 
+    // https://html.spec.whatwg.org/multipage/syntax.html#create-an-element-for-the-token
```
jdm: Remove `syntax.html`. However, we really will need comments labelling which steps are being implemented in the code, because it's really hard to figure out the association. I guess the majority of the new code is in support of the form owner stuff in step 12?

src/tree_builder/actions.rs
```diff
@@ -775,10 +743,39 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
...
+        let qname = QualName::new(ns, name);
+        let elem = self.sink.create_element(qname.clone(), attrs.clone());
+
+        let insertion_point = self.appropriate_place_for_insertion(None);
+        let tree_node = match insertion_point {
+            LastChild(ref p) |
+            BeforeSibling(ref p) => p.clone()
+        };
+
+        // Step 4.
```
jdm: This step number is wrong.

src/tree_builder/actions.rs
```diff
@@ -775,10 +743,39 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
...
+        // TODO: Handle template element case
+        if form_associatable(qname.clone())
+           && self.form_elem.is_some()
+           && !(reassociatable(qname.clone())
+               && attrs.iter().any(|a| a.name == qualname!("","form"))) {
```
jdm: nit: && belongs on the previous line in all these cases.

src/tree_builder/actions.rs
```diff
@@ -747,7 +747,7 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
     fn insert_element(&mut self, push: PushFlag, ns: Namespace, name: LocalName, attrs: Vec<Attribute>)
             -> Handle {
         declare_tag_set!(form_associatable =
-            "button" "fieldset" "input" "keygen" "label"
+            "button" "fieldset" "input" "keygen" "label" "legend"
```
jdm: https://html.spec.whatwg.org/multipage/forms.html#form-associated-element seems to contradict this.

src/tree_builder/actions.rs
```diff
@@ -775,10 +743,39 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
...
+        let tree_node = match insertion_point {
+            LastChild(ref p) |
+            BeforeSibling(ref p) => p.clone()
+        };
+
+        // Step 4.
+        // TODO: Handle template element case
+        if form_associatable(qname.clone())
+           && self.form_elem.is_some()
+           && !(reassociatable(qname.clone())
```
jdm: It's not clear why we're using reassociatable instead of the definition of [listed](https://html.spec.whatwg.org/multipage/forms.html#category-listed) here.

src/tree_builder/actions.rs
canaltinova: Hm, legend element's tests were not passing without this change. But maybe I missed something I'll look into that.

src/tree_builder/actions.rs
canaltinova: I think its name is changed since 2015. They weren't different except names. I renamed it that way.

src/tree_builder/actions.rs
```diff
@@ -775,10 +743,40 @@ impl<Handle, Sink> TreeBuilderActions<Handle>
...
+        let elem = self.sink.create_element(qname.clone(), attrs.clone());
+
+        let insertion_point = self.appropriate_place_for_insertion(None);
+        let tree_node = match insertion_point {
+            LastChild(ref p) |
+            BeforeSibling(ref p) => p.clone()
+        };
+
+        // Step 12.
+        // TODO: Handle template element case
```
jdm: What's required to handle this? We support template elements elsewhere in the parser now.

canaltinova: Yes, you're right. I forgot about it. Implemented it now, but I need to be sure that it's right :) https://html.spec.whatwg.org/multipage/syntax.html#create-an-element-for-the-token step 12 says:
> ...and there is no template element on the stack of open elements...

I hope I understand it right.

jdm:
If you squash the commits, I'll merge them. Thanks!

canaltinova:
Squashed it. Thanks for the help!

jdm:
We're going to need to make a new PR to merge this when https://github.com/servo/servo/issues/15283 is ready to merge.

nox:
@canaltinova Was there really no way to avoid separating the concept of appropriate place for insertion and the insertion itself?

PR #253, opened 2017-02-21
Remove incorrect `unsafe impl Send for Token { }`

SimonSapin:
Token contains StrTendril, which is not Send because it contains a shared reference to a `Cell<usize>` for reference-counting.

Refactor test harness to not need Send.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/253)


SimonSapin:
Let’s batch this with the next breaking change…

PR #256, opened 2017-03-04
Add "contents" and "scroll-position" keywords to local names

canaltinova:
They are needed for https://github.com/servo/servo/pull/15813

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/256)

PR #257, opened 2017-03-09
Revert "Add "contents" and "scroll-position" keywords to local names"

SimonSapin:
This reverts PR #256 / commit 6089ea41ad7acae12913d5c8b85864969da8fe5f.

Turns out these atoms are needed in `servo_atoms::Atom`, not `LocalName`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/257)

### DIFF
--- a/atoms/Cargo.toml
+++ b/atoms/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "html5ever-atoms"
-version = "0.2.1"
+version = "0.2.0"
 authors = [ "The html5ever Project Developers" ]
 license = "MIT / Apache-2.0"
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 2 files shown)


SimonSapin:
I’ve realized after submitting this that removing static atoms is technically a breaking change, and v0.2.1 is already on crates.io with them. @nox, what approach would you prefer?

* Keep the useless static atoms (revert this revert)
* Yank 0.2.1 (it was published 4 days ago and I highly doubt anyone depends on the new static atoms) and pretend it never happened
* Move to 0.3.0, which means h5e also gets a breaking release.

KiChjang:
I think yanking would be the correct approach here.

SimonSapin:
I tend to agree. Yanked. (We can still unyank or/and do 0.3.0.)

nox:
@SimonSapin I think we are supposed to release a 0.2.2 without the change that we yanked, otherwise even crates.io is confused.

SimonSapin:
Done.

PR #264, opened 2017-04-07
Update string-cache in html5ever-atoms.

emilio:
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/264)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "html5ever"
-version = "0.14.1"
+version = "0.15.0"
 authors = [ "The html5ever Project Developers" ]
 license = "MIT / Apache-2.0"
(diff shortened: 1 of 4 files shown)


emilio:
(Let me know if this should be breaking, but I don't think so)

nox:
It's a breaking change because `string_cache::Atom` is exposed.

nox:
So it's a breaking change in `html5ever` and `xml5ever` too.

emilio:
Sure, but its API isn't changed, right? Oh, I guess you could have another older string_cache or something like that/

nox:
Yes, using a type from a dependency with a different breaking version than before is a change in API, because `Atom` isn't the same `Atom`.

emilio:
Ammended

PR #268, opened 2017-04-29
Xhtml5ever

### LINKED ISSUE #210: QualName should be re-exported
The `QualName` type is used as part of the API, but it is pulled from the `string_cache` crate.
`QualName` should be reexported with a `pub use`, just like `Tendril` already is.

### LINKED ISSUE #261: html and xml named entity tables should be shared
Right now, we have `data/entities.json` and `xml5ever/data/entities.json`, which are identical.  If a project (such as Servo) requires both `html5ever` and `xml5ever`, it'll have duplicate copies of the named entity tables, which are not particularly small--about 230K on x86-64 Linux, roughly half that on 32-bit platforms.

Sticking the entities and the process to generate a phf table for them in a separate, shared crate would alleviate this problem.  There might be issues down the road if sets of entities in HTML and XML are no longer identical...but I'm sure solutions can be worked out there if necessary.

### LINKED ISSUE #266: Share more code with xml5ever
Yes, after discussion in https://github.com/Ygg01/xml5ever/issues/30 we’ve moved the xml5ever code into this repository with the intention to share more code. However much of that has yet to happen. Rough plan:

* Make every crate in the repo use a single [workspace](http://doc.crates.io/manifest.html#the-workspace-section)
* Make sure Travis-CI is running every test
* Rename the the `html5ever_atoms` crate to `markup5ever` and update html5ever and xml5ever to use it. This is a breaking change, so increment version numbers accordingly.
* Make it so that users of either html5ever or xml5ever don’t need to have an explicit dependency to markup5ever (which should be an implementation detail). One way to do this is to have `pub use markup5ever::*;` at the root of each crate.
* Find things that exist in both html5ever and xml5ever, and move them to markup5ever. Judge on a case-by-case basis when there are differences.

  For example `TokenSink` has some methods in common, but both parsers have some methods unique to it in their respective `TokenSink`. Since this trait is implemented by users of the libraries, the shared `TokenSink` should have the union of all methods (to support both parsers) and provide default implementations of methods where that makes sense (to reduce the overhead for users that only care about one parser).

  The `Parser` and `BytesParser` types in `html5ever::driver` are busted for handling `</script>`, we might want to remove them or redesign them. (Don’t bother adding it to xhtml5ever.)

Ygg01:
Ok, this is a large one.

Fixes #266, fixes #261, fixes #210.

It moves html5ever into separate folder, renames html5ever macros markup5ever and stores common code there. 

Here is short summary of what I know is and isn't done.

- [x] Make every crate in the repo use a single workspace
- [x] Make sure Travis-CI is running every test
- [x] Rename the the html5ever_atoms crate to markup5ever and update html5ever and xml5ever to use it.
- [x] Increment version numbers
- [x] Make it so that users of either html5ever or xml5ever don’t need to have an explicit dependency to markup5ever 
- [x] Export QualName #210 
- [x] let markup5ever generate entities.json #261    
- [x] **Move TokenSink to markup5ever**
- [x] Move TreeSink to markup5ever 
- [x] Move BufferQueue to markup5ever 
- [x] Move SmallCharSet to markup5ever 
- [x] **Deal with driver.rs**

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/268)


Ygg01:
Possible issues:

- better organization of pub use.
- `elem_name` and `elem_name_ref`  (and `same node`/`same_node_ref`) might be redundant. But I didn't want to change `elem_name` to use references, because possible unseen issues
- the code could use CargoFmt and Clippy

SimonSapin:
Wow, thanks a lot! I’ll look at the diffs later but based just on the commit messages this looks great.

A couple notes:

> * Move TokenSink to markup5ever

As I’m mentioned I don’t think this is important. When an application wants to support both XML and HTML it will most likely use each parser as a whole, not the tokenizer or the tree builder individually. So while it is useful to make `TreeSink` a shared/compatible interface (because that’s the output of the parser as a whole), `TokenSink` on the other hand is an internal implementation detail that is irrelevant to such an application.

> * Deal with driver.rs

There is stuff to do here, but I think it’s mostly independent of what’s in this PR at this point. So don’t worry about it :)

> ```diff
> -// Copyright 2014 The html5ever Project Developers. See the
> +// Copyright 2016 The html5ever Project Developers. See the
> ```

I think this should be a range: `Copyright 2014-2017`.

Ygg01:
My grep-fu failed me on first attempt. I had to rebase some Copyright fixes. I apologize for the inconvenience.

Thank you for kind words. I hope reviewing it won't be too difficult. BTW I won't be near computer for a few days, so I might take longer to respond/fix.

SimonSapin:
> BTW I won't be near computer for a few days, so I might take longer to respond/fix.

That’s fine. We’re not in a hurry with this, and this repository doesn’t see much activity at the moment so this will hopefully not bitrot.

However, if you prefer, I can take over the PR and fix up some minor things and land it in your absence. In that case I have one question (so far):

In `scripts/travis-build.sh`:
```sh
# This is a hack until cargo doc lands on stable
cargo doc -p html5ever
cargo doc -p xml5ever
```

What do you mean, “until cargo doc lands on stable”?

Ygg01:
Yeah, I forgot to fix that comment. I meant a hack until `cargo doc --all` works on stable.  Right now its nightly, I believe?

SimonSapin:
Re `cargo doc --all`, ok.

---

Reviewed 94 of 98 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, 9 of 9 files at r6, 13 of 13 files at r7, 14 of 14 files at r8, 11 of 12 files at r9, 8 of 8 files at r10, 5 of 5 files at r11, 26 of 26 files at r12, 10 of 10 files at r13, 7 of 7 files at r14, 56 of 56 files at r15.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[markup5ever/Cargo.toml, line 27 at r8](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuYDl21T4SNbZVgH2b:-KiuYDl21T4SNbZVgH2c:b8w4jzk) ([raw file](https://github.com/servo/html5ever/blob/3406fc154c013ac8e5ef06b23a7ddfe2df8f7016/markup5ever/Cargo.toml#L27)):*
> ```TOML
> phf_codegen = "0.7.3"
> rustc-serialize = "0.3.15"
> syn = { version = "0.11", features = ["full", "visit"] }
> ```

It looks like `syn` isn’t used in m5e?

---

*[markup5ever/lib.rs, line 20 at r12](https://reviewable.io:443/reviews/servo/html5ever/268#-KiubKi2UAssu59DuzN8:-KiubKi3bBECkjfqSl3U:b-hwtx5y) ([raw file](https://github.com/servo/html5ever/blob/19c89ff9b1cb5336b5967285d2d40b6bd4cfa739/markup5ever/lib.rs#L20)):*
> ```Rust
> macro_rules! qualname {
>     ("", $local:tt) => {
>         ::markup5ever::QualName {
> ```

Why not `$crate` here?

---

*[scripts/travis-build.sh, line 25 at r10](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuZau2NTgSk73QayJb:-KiuZau3fDdUrYCh2iKB:bqya0ty) ([raw file](https://github.com/servo/html5ever/blob/efc5d2210f956a2972b26fcb170c07b9984ba9d9/scripts/travis-build.sh#L25)):*
> ```Shell
> fi
> 
> # This is a hack until cargo doc lands on stable
> ```

What do you mean, “until cargo doc lands on stable”?

---

*[xml5ever/Cargo.toml, line 21 at r4](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuVCl8tUYA7ZQlzteJ:-KiuVCl9OEWHPsmQr3wa:b-r4d7r6) ([raw file](https://github.com/servo/html5ever/blob/a8a54c1a66736177b94d68c55c07625933b7bb40/xml5ever/Cargo.toml#L21)):*
> https://github.com/rust-lang/cargo/issues/1512

This issue is "plugins not working in doc tests". Since we’re not using compile plugins anymore this is no longer relevant. And the issue is fixed anyway.

If there is any doctest in either library we should enable them back. If not, we should keep `doctest = false` so that `cargo test` doesn’t spend time to collect zero doctests, but change the comment to say that instead.

---

*[xml5ever/src/rcdom.rs, line 43 at r12](https://reviewable.io:443/reviews/servo/html5ever/268#-KiubcdSCtOOwAANowbf:-KiubcdTUaPWFEO8DBGW:bytibmy) ([raw file](https://github.com/servo/html5ever/blob/19c89ff9b1cb5336b5967285d2d40b6bd4cfa739/xml5ever/src/rcdom.rs#L43)):*
> ```Rust
>     /// A script element and its "already started" flag.
>     /// https://html.spec.whatwg.org/multipage/#already-started
>     Script(bool),
> ```

While we’re touching this, let’s used a struct-like variant to name the field: `Script { already_started: bool }`

---

*[xml5ever/src/rcdom.rs, line 110 at r12](https://reviewable.io:443/reviews/servo/html5ever/268#-KiucEQUkPL-FqLfyMSH:-KiucEQUkPL-FqLfyMSI:b-185gex) ([raw file](https://github.com/servo/html5ever/blob/19c89ff9b1cb5336b5967285d2d40b6bd4cfa739/xml5ever/src/rcdom.rs#L110)):*
> ```Rust
> 
> #[allow(trivial_casts)]
> fn same_node(x: &Handle, y: &Handle) -> bool {
> ```

Let’s remove this and use `Rc::ptr_eq` instead. Unless we want to support older Rust versions? It’s added in 1.17.

---

*[xml5ever/tests/tokenizer.rs, line 1 at r5](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuWfEPyIVYgZDQ5Vcl:-KiuWfEQthYwHXAXMmAJ:b-dqtlh5) ([raw file](https://github.com/servo/html5ever/blob/a705bd6706c4f717c99375ab46f0cf1cd524cbcd/xml5ever/tests/tokenizer.rs#L1)):*
> ```Rust
> #![cfg_attr(feature = "unstable", feature(start, test))]
> ```

I think this can be removed as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/html5ever/268#-:-KiuezDtMm5_DZXdMFTe:baq07x7)*

Ygg01:
*[scripts/travis-build.sh, line 25 at r10](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuZau2NTgSk73QayJb:-Kium4RD2PZN4EEv5fl7:bv8bu50) ([raw file](https://github.com/servo/html5ever/blob/efc5d2210f956a2972b26fcb170c07b9984ba9d9/scripts/travis-build.sh#L25)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

What do you mean, “until cargo doc lands on stable”?
</blockquote></details>

It's an error on my part, caused by old rustc. `cargo test --all ` works on stable, but my older version didn't allow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/html5ever/268)*

Ygg01:
*[markup5ever/Cargo.toml, line 27 at r8](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuYDl21T4SNbZVgH2b:-KiunRkvJNNC-HVjpvVQ:bi8oyrv) ([raw file](https://github.com/servo/html5ever/blob/3406fc154c013ac8e5ef06b23a7ddfe2df8f7016/markup5ever/Cargo.toml#L27)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

It looks like `syn` isn’t used in m5e?
</blockquote></details>

Hm, I was under the impression, that it was used, since no warning was issued. But you are right.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/html5ever/268)*

Ygg01:
*[xml5ever/Cargo.toml, line 21 at r4](https://reviewable.io:443/reviews/servo/html5ever/268#-KiuVCl8tUYA7ZQlzteJ:-KiuoS8t85okAZceJkP4:bngg713) ([raw file](https://github.com/servo/html5ever/blob/a8a54c1a66736177b94d68c55c07625933b7bb40/xml5ever/Cargo.toml#L21)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

> https://github.com/rust-lang/cargo/issues/1512

This issue is "plugins not working in doc tests". Since we’re not using compile plugins anymore this is no longer relevant. And the issue is fixed anyway.

If there is any doctest in either library we should enable them back. If not, we should keep `doctest = false` so that `cargo test` doesn’t spend time to collect zero doctests, but change the comment to say that instead.
</blockquote></details>

Enable this in html5ever, as well I assume? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/html5ever/268)*

SimonSapin:
> Enable this in html5ever, as well I assume?

Both.

SimonSapin:
Reviewed 2 of 98 files at r1, 3 of 9 files at r6, 1 of 8 files at r10, 1 of 10 files at r13, 1 of 1 files at r16, 6 of 6 files at r17, 2 of 2 files at r18, 2 of 2 files at r19.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[markup5ever/interface/mod.rs, line 139 at r17](https://reviewable.io:443/reviews/servo/html5ever/268#-KivCd4LXcm-Tlz0gZnG:-KivCd4LXcm-Tlz0gZnH:b-ikr3cs) ([raw file](https://github.com/servo/html5ever/blob/a8c619c13aa2c54a75965443c71a437c0d14d7d1/markup5ever/interface/mod.rs#L139)):*
> ```Rust
>     #[test]
>     fn qualname_macro() {
>         assert_eq!(qualname!("", ""), QualName { ns: ns!(), local: local_name!(""), prefix: None });
> ```

`qualname!` is sometimes used as a pattern. I don’t know that we should be checking that the prefix is `None` in those case… We may want separate macros: one with a `prefix: _` pattern, one for expressions. I’ll survey usage of the macro, in both h/x5e and Servo.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/html5ever/268)*

Ygg01:
@SimonSapin. What to do about `qualname`?

SimonSapin:
This PR looks good, thanks! I added more stuff on top at: https://github.com/Ygg01/html5ever/pull/1. That includes taking care of `qualname!` (by removing most of its usage and making it private).

Ygg01:
Review status: 49 of 90 files reviewed at latest revision, 2 unresolved discussions.

---

*[xml5ever/src/rcdom.rs, line 110 at r12](https://reviewable.io:443/reviews/servo/html5ever/268#-KiucEQUkPL-FqLfyMSH:-KiuneFgayGCKgQB1C_E:bih9f5k) ([raw file](https://github.com/servo/html5ever/blob/19c89ff9b1cb5336b5967285d2d40b6bd4cfa739/xml5ever/src/rcdom.rs#L110)):*
<details><summary><i>Previously, SimonSapin (Simon Sapin) wrote…</i></summary><blockquote>

Let’s remove this and use `Rc::ptr_eq` instead. Unless we want to support older Rust versions? It’s added in 1.17.
</blockquote></details>

AFAIK no need for support. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/html5ever/268)*

Ygg01:
What part of review is blocking progress?

PR #277, opened 2017-06-15
change type ChildrenOnly to ChildrenOnly(Option<QualName>)

moonlightdrive:
This is necessary so that these nodes exist on the stack when we
serialize the children nodes and may need the parent for context.
Addresses servo/servo#14975

---
Opening this PR for feedback/review

~~This does work and solves the above bug. I'm not sure if this is the best implementation, as there's a bit of duplicated code. I thought about adding a boolean flag to `start_elem` that would determine whether the writes should be performed. But then I thought maybe when not writing, it's best not to call a function returning `io::Result<()>`?~~

Also not 100% sure if the xml implementation is the right thing to do

Anyway, please let me know of any improvements/changes to make. Thanks :)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/277)

### DIFF
--- a/html5ever/src/serialize/mod.rs
+++ b/html5ever/src/serialize/mod.rs
@@ -15,11 +15,11 @@ use {LocalName, QualName};
 
 pub fn serialize<Wr, T>(writer: Wr, node: &T, opts: SerializeOpts) -> io::Result<()>
 where Wr: Write, T: Serialize {
-    let mut ser = HtmlSerializer::new(writer, opts);
+    let mut ser = HtmlSerializer::new(writer, opts.clone());
     node.serialize(&mut ser, opts.traversal_scope)
 }
 
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 pub struct SerializeOpts {
     /// Is scripting enabled?
     pub scripting_enabled: bool,
@@ -39,7 +39,7 @@ impl Default for SerializeOpts {
     fn default() -> SerializeOpts {
         SerializeOpts {
             scripting_enabled: true,
-            traversal_scope: TraversalScope::ChildrenOnly,
+            traversal_scope: TraversalScope::ChildrenOnly(None),
             create_missing_parent: false,
         }
     }
@@ -72,10 +72,18 @@ fn tagname(name: &QualName) -> LocalName {
 
 impl<Wr: Write> HtmlSerializer<Wr> {
     fn new(writer: Wr, opts: SerializeOpts) -> Self {
+        let html_name = match opts.traversal_scope {
+            TraversalScope::IncludeNode | TraversalScope::ChildrenOnly(None) => None,
+            TraversalScope::ChildrenOnly(Some(ref n)) => Some(tagname(n))
+        };
         HtmlSerializer {
             writer: writer,
             opts: opts,
-            stack: vec![Default::default()],
+            stack: vec!(ElemInfo {
+                html_name: html_name,
+                ignore_children: false,
+                processed_first_child: false,
+            }),
         }
     }
 
@@ -190,18 +198,6 @@ impl<Wr: Write> Serializer for HtmlSerializer<Wr> {
     }
 
     fn write_text(&mut self, text: &str) -> io::Result<()> {
-        let prepend_lf = text.starts_with("\n") && {
-            let parent = self.parent();
-            !parent.processed_first_child && match parent.html_name {
-                Some(local_name!("pre")) | Some(local_name!("textarea")) | Some(local_name!("listing")) => true,
-                _ => false,
-            }
-        };
-
-        if prepend_lf {
-            try!(self.writer.write_all(b"\n"));
-        }
-
         let escape = match self.parent().html_name {
             Some(local_name!("style")) | Some(local_name!("script")) | Some(local_name!("xmp"))
             | Some(local_name!("iframe")) | Some(local_name!("noembed")) | Some(local_name!("noframes"))
--- a/html5ever/tests/serializer.rs
+++ b/html5ever/tests/serializer.rs
@@ -165,15 +165,15 @@ test!(noframes_literal, r#"<noframes>(x & 1) < 2; y > "foo" + 'bar'</noframes>"#
 
 test!(pre_lf_0, "<pre>foo bar</pre>");
 test!(pre_lf_1, "<pre>\nfoo bar</pre>", "<pre>foo bar</pre>");
-test!(pre_lf_2, "<pre>\n\nfoo bar</pre>");
+test!(pre_lf_2, "<pre>\n\nfoo bar</pre>", "<pre>\nfoo bar</pre>");
 
 test!(textarea_lf_0, "<textarea>foo bar</textarea>");
 test!(textarea_lf_1, "<textarea>\nfoo bar</textarea>", "<textarea>foo bar</textarea>");
-test!(textarea_lf_2, "<textarea>\n\nfoo bar</textarea>");
+test!(textarea_lf_2, "<textarea>\n\nfoo bar</textarea>", "<textarea>\nfoo bar</textarea>");
 
 test!(listing_lf_0, "<listing>foo bar</listing>");
 test!(listing_lf_1, "<listing>\nfoo bar</listing>", "<listing>foo bar</listing>");
-test!(listing_lf_2, "<listing>\n\nfoo bar</listing>");
+test!(listing_lf_2, "<listing>\n\nfoo bar</listing>", "<listing>\nfoo bar</listing>");
 
 test!(comment_1, r#"<p>hi <!--world--></p>"#);
 test!(comment_2, r#"<p>hi <!-- world--></p>"#);
--- a/markup5ever/rcdom.rs
+++ b/markup5ever/rcdom.rs
@@ -326,7 +326,7 @@ impl Default for RcDom {
 impl Serialize for Handle {
     fn serialize<S>(&self, serializer: &mut S, traversal_scope: TraversalScope) -> io::Result<()>
     where S: Serializer {
-        match (traversal_scope, &self.data) {
+        match (&traversal_scope, &self.data) {
             (_, &NodeData::Element { ref name, ref attrs, .. }) => {
                 if traversal_scope == IncludeNode {
                     try!(serializer.start_elem(name.clone(),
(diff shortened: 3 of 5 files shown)


Ygg01:
Shouldn't there be test, to check in case this feature regresses in future?

moonlightdrive:
Sure thing. How would I add tests? By adding to `html5ever/tests/serializer.rs`? And how would I run them? `cargo test` from `html5ever` directory?

Ygg01:
A fair question. I think we put them in [html5ever serializer tests](https://github.com/servo/html5ever/blob/5081b2cb151439686ce4e0d3a1956a73e8ca4026/html5ever/tests/serializer.rs).

Actually, if you can call `cargo test -p html5ever`, to run only h5e tests. See for examples [workspace tests](https://github.com/servo/html5ever/blob/5081b2cb151439686ce4e0d3a1956a73e8ca4026/scripts/travis-build.sh)

html5ever/src/serialize/mod.rs
```diff
@@ -94,6 +107,30 @@ impl<Wr: Write> HtmlSerializer<Wr> {
 }
 
 impl<Wr: Write> Serializer for HtmlSerializer<Wr> {
+    fn start_without_write(&mut self, name: QualName) {
```
jdm: I'm not sure I understand why this is an improvement over my suggested solution in https://github.com/servo/servo/issues/14975#issuecomment-307263635 . Am I right that the code that uses the serializer would choose to call start_without_write or start_elem with the parent node, based on whether ChildrenOnly is selected? This seems like it makes the serializer interface more complicated than necessary to me.

jdm:
Sorry for the long delay in providing feedback. Could you explain why you ended up choosing this implementation? There's a gap between my proposed solution and your comment suggesting this alternative one which doesn't make clear why it was necessary.

moonlightdrive:
No worries about the wait! Apologies for the incoming wall of text.

I actually don't think this is the right solution/implementation anymore. (Definitely opened this PR prematurely.) I think there are two solutions

1. Change the `TraversalScope` type to
``` rust
pub enum TraversalScope {
    IncludeNode,
    ChildrenOnly(QualName)
}
```

This way the parent node is accessible from Serializer::new, or anywhere else in html5ever, and can be pushed to the stack as we discussed.

But I haven't gotten this to work due to `QualName` deriving `Copy`; [QualName struct](https://github.com/servo/html5ever/blob/fd514b930d4b0af131a3838cb304acd0b4664a88/markup5ever/interface/mod.rs#L106) contains an `Option<Prefix>` field. [Prefix](https://github.com/servo/html5ever/blob/master/markup5ever/build.rs#L53), which does not derive the `Copy` trait, seems to be defined using the string-cache library. I'm stuck on using this to get `Prefix` to derive traits.

I think having `Prefix` derive `Copy` might be undesirable? My understanding is the purpose of a string interning library is to not have duplicate strings. If that's what's going on here? I don't understand the benefit of defining `Prefix` through this library.


2. Change [`start_elem`](https://github.com/servo/html5ever/blob/master/markup5ever/serialize.rs#L26) and `end_elem` to take a bool that determines whether the node should actually be serialized (popping/pushing the node regardless of the bool's value)

```rust
start_elem(&mut self, name: QualName, attrs:AtterIter, serialize_elem: bool) -> io::Result<()>
```

[Instead of](https://github.com/servo/servo/blob/ed7686b42c02c5c1d769be952ef58b285f35a7f6/components/script/dom/servoparser/html.rs#L127) calling `start_elem`/`end_elem` from servo only when `traversalScope == IncludeNode`, we would always call `start_elem`/`end_elem` and pass `true` in the `IncludeNode` case, `false` otherwise.

I do have this implementation working locally. It's different than what we discussed & I'm not sure if it's logically the right thing to do.

jdm:
@moonlightdrive It sounds like the problem with my solution is that TraversalScope currently [derives Copy](https://github.com/servo/html5ever/blob/cee43bd304b7bfaadd7c25ed0c53caeebfadfe54/markup5ever/serialize.rs#L14), and since `QualName` does not (and cannot) implement Copy, this is an error. In that case, the answer is to remove `Copy` from the list of derived traits on TraversalScope.

moonlightdrive:
I assumed that the trait was derived for a particular reason and that removing it would be undesirable. Cursory google search has now led me to believe that it's standard Rust practice to derive Copy unless there's a reason not to? So I will go ahead and remove Copy and see where that leads me. Thanks @jdm for your response :)

moonlightdrive:
I seem to have broken some wpt tests related to this bit of code. innerHTML of `<pre>\x0A</pre>` should serialize to `\n` but after my change is serializing to `'\n\n'`. This seems to be [the relevant bit of code](https://github.com/servo/html5ever/blob/master/html5ever/src/serialize/mod.rs#L173:L179)

[Spec](https://www.w3.org/TR/2009/WD-html5-20090212/serializing-html-fragments.html) says, when current node is an element
> If current node is a `pre`, `textarea`, or `listing` element, append a U+000A LINE FEED (LF) character.)

Should this be done in [start_elem](https://github.com/servo/html5ever/blob/master/html5ever/src/serialize/mod.rs#L97) instead of `write_text`?

jrozner:
Is there a timeline for merging this into mainline or a plan to even do it at this point. We're depending on this fork currently for software we're writing that has to deal with processing malformed html token streams where close and open tags don't necessarily match up. We'd like to get to mainline and help push this over the finish line in whatever way we can.

jdm:
@jrozner What are you referring to, exactly? This PR is addressing a problem with serializing fragments; it has nothing to do with parsing.

jrozner:
@jdm An earlier commit we had been working with used the start_without_write methods introduced earlier to deal with serializing out invalid html without fixing it after running this parser. It isn't a parsing issue but and issue with the serializer not being able to handle malformed html which is what we need in our case. We are doing the fixup ourself without running the parser just by referencing the stack.

jechas01:
@jdm Just to further clarify what @jrozner's saying, our flow looks like this:

Potentially malformed hmtl input -> html5ever lexer -> process token stream -> serialize with the html5ever serializer

We'd like to be able to run the serializer without having to fabricate start tokens.

jdm:
Ok, that helps me understand a bit better. This is currently waiting on me to review the most recent changes; the recent lack of activity was waiting on addressing previous review comments. The goal is to address the underlying issue for servo/servo#14975; it's not clear to me if that's the same problem that you're using these changes to solve.

jechas01:
No, not quite the same problem that we were using this change to solve (though @jrozner informs me that we see that issue as well), but the stated rationale:

> This is necessary so that these nodes exist on the stack when we
serialize the children nodes and may need the parent for context.

was exactly what I was searching for when I came across this.

If the future of this PR doesn't quite align with our goals, maybe a different one would be needed?

jdm:
I'm open to other changes to address your use case, but a unit test demonstrating your issue would go a long way to clearing up what we're actually discussing.

jdm:
(in a separate issue, please)

html5ever/src/serialize/mod.rs
```diff
@@ -63,11 +63,15 @@ fn tagname(name: &QualName) -> LocalName {
 
 impl<Wr: Write> HtmlSerializer<Wr> {
     fn new(writer: Wr, opts: SerializeOpts) -> Self {
+        let qname = match &opts.traversal_scope {
```
jdm: The & in the match should not be necessary.

html5ever/src/serialize/mod.rs
```diff
@@ -63,11 +63,15 @@ fn tagname(name: &QualName) -> LocalName {
...
+            &TraversalScope::IncludeNode | &TraversalScope::ChildrenOnly(None) => None,
+            &TraversalScope::ChildrenOnly(Some(ref n)) => Some(n.clone())
```
jdm: Let's do `Some(tagname(n))` instead.

markup5ever/rcdom.rs
```diff
@@ -326,7 +326,7 @@ impl Default for RcDom {
 impl Serialize for Handle {
     fn serialize<S>(&self, serializer: &mut S, traversal_scope: TraversalScope) -> io::Result<()>
     where S: Serializer {
-        match (traversal_scope, &self.data) {
+        match (traversal_scope.clone(), &self.data) {
```
jdm: Let's do `(&traversal_scope, &self.data)` instead.

jdm:
Whoops, I totally missed https://github.com/servo/html5ever/pull/277#issuecomment-317278071 . That's an interesting result! The specification we should refer to is https://html.spec.whatwg.org/multipage/parsing.html#serialising-html-fragments ; interestingly, I do not see any equivalent for "If current node is a pre, textarea, or listing element, append a U+000A LINE FEED (LF) character." in that specification.

jdm:
In fact, further down the specification I see:
```
For historical reasons, this algorithm does not round-trip an initial U+000A LINE FEED (LF) character in pre, textarea, or listing elements, even though (in the first two cases) the markup being round-tripped can be conforming. The HTML parser will drop such a character during parsing, but this algorithm does not serialize an extra U+000A LINE FEED (LF) character.
```
I think this means we should remove that bit from the implementation entirely!

moonlightdrive:
Thanks for the review @jdm. I made the changes; is this looking better?

jdm:
Some automated tests will need to be updated to account for the change:
```
...................thread 'listing_lf_2' panicked at 'assertion failed: `(left == right)`

  left: `"<listing>\n\nfoo bar</listing>"`,

 right: `"<listing>\nfoo bar</listing>"`', html5ever/tests/serializer.rs:176:0

note: Run with `RUST_BACKTRACE=1` for a backtrace.

F......thread 'pre_lf_2' panicked at 'assertion failed: `(left == right)`

  left: `"<pre>\n\nfoo bar</pre>"`,

 right: `"<pre>\nfoo bar</pre>"`', html5ever/tests/serializer.rs:168:0

F............thread 'textarea_lf_2' panicked at 'assertion failed: `(left == right)`

  left: `"<textarea>\n\nfoo bar</textarea>"`,

 right: `"<textarea>\nfoo bar</textarea>"`', html5ever/tests/serializer.rs:172:0

F.

failures:

failures:

    listing_lf_2

    pre_lf_2

    textarea_lf_2
```

jdm:
Otherwise this looks good!

html5ever/tests/serializer.rs
```diff
@@ -165,15 +165,15 @@ test!(noframes_literal, r#"<noframes>(x & 1) < 2; y > "foo" + 'bar'</noframes>"#
 
 test!(pre_lf_0, "<pre>foo bar</pre>");
 test!(pre_lf_1, "<pre>\nfoo bar</pre>", "<pre>foo bar</pre>");
-test!(pre_lf_2, "<pre>\n\nfoo bar</pre>");
+test!(pre_lf_2, "<pre>foo bar</pre>");
```
jdm: As written, this is identical to pre_lf_0. Instead, we should do `test!(pre_lf_2, "<pre>\n\nfoo bar</pre>", "<pre>\n\nfoo bar</pre>");` (and same for the other modified tests).

moonlightdrive: oh whoops.. my mistake.. not sure what was going on in my mind :|

---

Should it actually be `test!(pre_lf_2, "<pre>\n\nfoo bar</pre>", "<pre>\nfoo bar</pre>");`? 

> For example, consider the following markup:
>``` 
><pre>
>
>Hello.</pre>
>```
>
>When this document is first parsed, the pre element's child text content starts with a single newline character. After a serialize-reparse roundtrip, the pre element's child text content is simply "Hello.".

I'm don't understand what's meant by a serialize roundtrip.. but `test!(pre_lf_2, "<pre>\n\nfoo bar</pre>", "<pre>foo bar</pre>");` doesn't seem to be right

jdm: Ah, yes, you're right. I shouldn't have tried to do that from memory :)

PR #284, opened 2017-06-29
"unstable" Cargo features are now no-ops, to be removed.

SimonSapin:
https://github.com/servo/tendril/pull/33

This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/284)

### DIFF
--- a/html5ever/Cargo.toml
+++ b/html5ever/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "html5ever"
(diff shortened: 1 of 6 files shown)


SimonSapin:
New version of `rustc-test` should fix this.

@bors-servo retry

SimonSapin:
Newer version of `rustc-test` should fix this more.

@bors-servo retry

PR #291, opened 2017-08-01
reexport string_cache

LuoZijun:
reexport string_cache

### DIFF
--- a/markup5ever/lib.rs
+++ b/markup5ever/lib.rs
@@ -9,7 +9,7 @@
 
 #[cfg(feature = "heap_size")] #[macro_use] extern crate heapsize_derive;
 #[cfg(feature = "heap_size")] extern crate heapsize;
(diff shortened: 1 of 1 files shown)


SimonSapin:
Out of curiosity: why? Do you want to use string-cache to define more types yourself, or is it also needed somehow to use with the existing `LocalName`, `Namespace`, and `Prefix` types?

LuoZijun:
@SimonSapin yes, like this : https://github.com/tbodt/htmlpyever/blob/master/glue.rs#L355

SimonSapin:
Does `LocalName::from` not work?

LuoZijun:
it work. sry, my mistake.

but i can't not find interface ( `LocalName::from` ) in this document page: https://doc.servo.org/markup5ever/type.LocalName.html

PR #293, opened 2017-08-02
Don't bail on empty Serializer stack

### LINKED ISSUE #292: Feature request: Ability to serialize malformed token stream
We have a use case that requires the ability to re-serialize malformed html input without needing to insert extra open tags to appease the Serializer, which panics when it tries to serialize a close tag when the stack is empty.

The obvious (naive perhaps?) solution is to provide a mechanism to manipulate the stack in order to fake the presence of an open tag. Alternatively, the behavior could be changed in the `parent` and `end_elem` methods to not panic in the absence of an `ElemInfo` on the stack, and to instead either create or return a sane default. I'll start working on it from that angle, but I'm open to alternative approaches.

jechas01:
This prevents panics if `end_elem` or `parent` is called with an empty stack by fabricating a new default ElemInfo.

Normally, this will never happen with valid html, but in the event that someone needs to serialize a token stream representing a malformed tree, they can now do so.

Fixes #292

### DIFF
--- a/html5ever/src/serialize/mod.rs
+++ b/html5ever/src/serialize/mod.rs
@@ -26,6 +26,13 @@ pub struct SerializeOpts {
 
     /// Serialize the root node? Default: ChildrenOnly
     pub traversal_scope: TraversalScope,
+
+    /// If the serializer is asked to serialize an invalid tree, the default
+    /// behavior is to panic in the event that an `end_elem` is created without a
+    /// matching `start_elem`. Setting this to true will prevent those panics by
+    /// creating a default parent on the element stack. No extra start elem will
+    /// actually be written. Default: false
+    pub create_missing_parent: bool,
 }
 
 impl Default for SerializeOpts {
@@ -33,6 +40,7 @@ impl Default for SerializeOpts {
         SerializeOpts {
             scripting_enabled: true,
             traversal_scope: TraversalScope::ChildrenOnly,
+            create_missing_parent: false,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


jdm:
One thing that would help me understand is a unit test that fails before these changes and passes afterwards.

jechas01:
Whoops, must have messed something up when I reverted an accidental push. Still working on a test case.

jechas01:
OK, now the `malformed_tokens` test added in 20b04ab will fail with `thread 'malformed_tokens' panicked at 'no parent ElemInfo', /checkout/src/libcore/option.rs:819:4`
without 0978968.

html5ever/src/serialize/mod.rs
```diff
@@ -159,7 +160,10 @@ impl<Wr: Write> Serializer for HtmlSerializer<Wr> {
     }
 
     fn end_elem(&mut self, name: QualName) -> io::Result<()> {
-        let info = self.stack.pop().expect("no ElemInfo");
+        let info = self.stack.pop().unwrap_or_else(|| {
```
jdm: Rather than unconditionally creating a missing element, I would prefer to add an option to SerializeOpts and only create the parent if this behaviour is specifically requested.

html5ever/tests/serializer.rs
```diff
@@ -28,33 +102,65 @@ fn parse_and_serialize(input: StrTendril) -> StrTendril {
...
+}
+
+
+
 test!(empty, r#""#);
 test!(smoke_test, r#"<p><i>Hello</i>, World!</p>"#);
 
-test!(misnest, r#"<p><i>Hello!</p>, World!</i>"#,
-    r#"<p><i>Hello!</i></p><i>, World!</i>"#);
+test!(
```
jdm: Are these changes caused by running rustfmt? I would prefer to split them into a separate commit or revert them entirely.

jechas01: Ah, yeah. Thought I `git add -p`'d around them, but I guess not. Should be an easy change.

jdm:
@bors-servo: r+
Thanks!

PR #295, opened 2017-08-03
Add srcset local name for "srcset parsing algorithm"

iamneha:


jdm:
Could you change https://github.com/servo/html5ever/blob/master/markup5ever/Cargo.toml#L3 to 0.3.2 as well?

Issue #289, opened 2017-07-30
Guarantee O(n) time even for malicious input


DemiMarie:
HTML parsers are not just used in client-side applications: they are also used on servers, such as in HTML sanitizers.  html5ever (and xml5ever) should guarantee that they cannot be coerced into taking more than O(n), or at worst O(n log n), time.  This may be difficult, especially if one does not want to use massively connected datastructures. 

Right now, it seems that the worst offenders are likely to be calls to `Vec::remove` in various places, such as in the adoption agency algorithm.

Ygg01:
Excellent point.  I think attribute duplication is probably the worst offender, especially in xml5ever, since it does attribute removal twice (first time in tokenizer and second after namespace resolution).

Any ideas how other parsers limit the execution time to `O(n)` (or `O(n log n)` )?

DemiMarie:
I know Gumbo does worse than html5ever; really deep nesting is enough to
make it go quadratic.

The main problem seems to be the active formatting elements and unclosed
elements vectors.  I am not sure how to pull off guaranteed linear time.
Maybe by caching, or by storing flags instead of searching the lists?  Some
steps seem to require a linked list be used instead of a vector, so as to
allow O(1) removal from the middle of the list.  Others seem to require a
hash set of every node, keyed by name, attributes, and values, that allows
up to 3 duplicates!  I think that using vectors is a losing battle here,
which strongly favors using an arena for all allocations inside the parser.

DemiMarie:
Another option is to deliberately impose an artificial cap on the length of
those lists.  Ugly, and in violation of the spec, but the only simple
option I see.

SimonSapin:
I’ve changed the title because "guarantee" is not gonna happen unless we have static analysis that can tell if we’re doing something wrong, and that sounds like a hard research problem.

Do you know specific inputs where html5ever currently behaves badly?

DemiMarie:
A long string of opening <I> tags, followed by a bunch of nested links.
Causes quadratic time consumption.

DemiMarie:
@SimonSapin A long string of opening `<i>` tags, followed by a bunch of nested links, causes quadratic time consumption in the adoption agency algorithm.  You can reproduce this with `perl -e 'my $x = 20000; print("<i>"x$x . "<a>"x$x . "</a>"x$x)' | html2html`.

DemiMarie:
So does

~~~sh
[...]

PR #274, opened 2017-06-01
Remove [Xml]TreeBuilderActions and [Xml]TreeBuilderStep and move their implementations into [Xml]TreeBuilder

jbg:
See issue #258

---
This change is [Reviewable](https://reviewable.io/reviews/servo/html5ever/274)

PR #297, opened 2017-08-06
Adoption agency limit

DemiMarie:
This hardens the code against denial of service attacks by only going back a certain number of elements (set at parser construction time) in the list of active formatting elements and the stack of open elements.  With this change, I have not been able to find any ways to cause html5ever to hang, even on pathological input.

This is a breaking change, since it adds an element to `TreeBuilderOpts`.


DemiMarie:
My understanding is that it is impossible to solve #289 without imposing *some* arbitrary limit – otherwise, it is possible to produce DOMs that are quadratic size.

Ygg01:
@DemiMarie #274 moved implentations from html5ever/src/tree_builder/actions.rs into html5ever/src/tree_builder/mod.rs, which broke this change. You will need to rebase.

DemiMarie:
@Ygg01 Rebased and squashed.

I went back to a limit of 10, because I doubt that most web pages will hit that limit, and because the limit*8 is roughly the “amplification factor” that a malicious document can achieve.  This does cause some tests to fail though.

Part of this can be solved by better algorithms, but part of it is inherent to (and IMO a bug in) the HTML parsing algorithm, which allows for quadratically-sized DOMs to be created from linearly-sized input.

Ygg01:
Hm. It seems your code is causing some errors in tree building. Or perhaps testing requires larger limit? I'll inspect once I'm home.

DemiMarie:
Testing requires a larger limit.  I chose the limit I did because the code
is already slow and I didn't want for an attacker to be able to DoS a
server.

I think that what is really needed is a better data structure.  At some
point, PRs should be issued against the spec that replace the algorithm
with an equivalent but faster version.

On Aug 11, 2017 11:10 AM, "Ygg01" <notifications@github.com> wrote:

Hm. It seems your code is causing some errors in tree building. Or perhaps
testing requires larger limit? I'll inspect once I'm home.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/servo/html5ever/pull/297#issuecomment-321839903>, or mute
the thread

Ygg01:
> I think that what is really needed is a better data structure. At some point,

Are there any examples what said data structure is?

DemiMarie:
@Ygg01 Testing requires a larger limit.  I chose the limit I did because the code is already slow and I didn't want for an attacker to be able to DoS a server.

I think that what is really needed is a better data structure.  At some point, PRs should be issued against the spec that replace the algorithm with an equivalent but faster version.  The algorithm in the spec should be one that people can actually use securely.

DemiMarie:
- Use some sort of hash table for the list of active formatting elements,
to allow keeping track of the “max. 3 duplicates” rule.
- Keep flags that can be checked, and which are updated whenever something
is added or removed, to avoid traversing lists a bunch.

usamec:
Any updates on this PR? 
(having HTML parser, which does not blow on pathological pages would be really nice).

SimonSapin:
@hsivonen, what does Gecko do here?

diejmon:
Tried this PR with the latest version and it doesn't fix stack overflow issue.

hsivonen:
I'm so bad at GitHub notifications. Sorry.

@SimonSapin, it's a bit unclear to me what "here" means. The [adoption agency algorithm](https://html.spec.whatwg.org/#adoption-agency-algorithm) has its outer loop count limited to 8 in the spec, and that's what Gecko limits it to.

Additionally, there's a need to limit the general DOM depth. On that point, you shouldn't copy Gecko but Blink. Please take a look at the [Gecko patch to match Blink](https://reviewboard.mozilla.org/r/178790/diff/21#index_header). It hasn't landed due to the sad combination of Android runtime stack limits and Robohornet. I'm hoping to resolve the Android issue one way or another in 2018H2.

Ygg01:
@hsivonen Could you explain the gist behind the Blink's limit, in that patch, My C++ foo isn't great. From what I could tell. It creates a limited `Vec` or `array` that has up to 512 (or rather 511) elements and errors and returns last element of said array.

hsivonen:
> Could you explain the gist behind the Blink's limit, in that patch, My C++ foo isn't great.

ReviewBoard went away, but [here is rebased version of the Gecko patch](https://bugzilla.mozilla.org/attachment.cgi?id=9031216&action=diff).

In places where the code calls `nodeFromStackWithBlinkCompat()`, instead of taking the most-recently-pushed element from the tree builder stack, element at index 511 is taken if there are more elements on the stack. Assumes zero-based indexing where the "html" element is at index 0.

Ygg01:
@DemiMarie could you work in the changes hsivonen mentioned?

DemiMarie:
@hsivonen I am working on it.

hsivonen:
The fix has finally landed in Gecko and is available for testing in Firefox Nightly.

freddyb:
@DemiMarie Are you stil working on this?

DemiMarie:
I am not.

PR #300, opened 2017-08-12
TreeBuilder no longer relies on `same_tree` and `has_parent_node`

cynicaldevil:
The use of `same_tree` and `has_parent_node` meant that the TreeBuilder needed access to the DOM state to carry out its tasks.


cynicaldevil:
cc @nox

PR #308, opened 2017-09-06
Removed has_parent_node and same_tree methods from TreeSink trait

### LINKED ISSUE #303: Remove `same_tree` and `has_parent_node` methods from the TreeSink trait
These methods are no longer used by the TreeBuilder and can be removed safely.

rozuur:
fixes #303

### DIFF
--- a/markup5ever/rcdom.rs
+++ b/markup5ever/rcdom.rs
@@ -207,13 +207,6 @@ impl TreeSink for RcDom {
         Node::new(NodeData::ProcessingInstruction { target: target, contents: data })
     }
 
-    fn has_parent_node(&self, node: &Handle) -> bool {
-        let parent = node.parent.take();
-        let has_parent = parent.is_some();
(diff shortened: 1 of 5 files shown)


markup5ever/rcdom.rs
```diff
@@ -270,7 +263,11 @@ impl TreeSink for RcDom {
         prev_element: &Self::Handle,
         child: NodeOrText<Self::Handle>) {
 
-        if self.has_parent_node(element) {
+        let parent = element.parent.take();
+        let has_parent = parent.is_some();
+        element.parent.set(parent);
+
+        if has_parent {
```
cynicaldevil:
```rust
if element.get().map_or(false, |p| p.upgrade().is_some())
```
 I think this should work.

Ygg01:
Small nitpick, you might want to update xml5ever to reflect the missing methods, since it uses markup5ever.

PR #312, opened 2017-09-17
Pass intended parent to create_element

cbrewster:
This is needed to resolve servo/servo#18277

### DIFF
--- a/html5ever/src/tree_builder/mod.rs
+++ b/html5ever/src/tree_builder/mod.rs
@@ -12,7 +12,7 @@
 //! The HTML5 tree builder.
 
 pub use interface::{QuirksMode, Quirks, LimitedQuirks, NoQuirks};
-pub use interface::{NodeOrText, AppendNode, AppendText, Attribute};
+pub use interface::{NodeOrText, AppendNode, AppendText, Attribute, IntendedParent};
 pub use interface::{TreeSink, Tracer, NextParserState, create_element, ElementFlags};
 
 use self::types::*;
@@ -741,7 +741,7 @@ impl<Handle, Sink> TreeBuilder<Handle, Sink>
                 // own, once as part of t.clone() above)?
                 let new_element = create_element(
                     &mut self.sink, QualName::new(None, ns!(html), tag.name.clone()),
-                    tag.attrs.clone());
+                    tag.attrs.clone(), IntendedParent::None);
                 self.open_elems[node_index] = new_element.clone();
                 self.active_formatting[node_formatting_index] = Element(new_element.clone(), tag);
                 node = new_element;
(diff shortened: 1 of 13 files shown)


KiChjang:
I sort of believe that the intended parent parameter shouldn't be optional, since the spec did not mention that it is optional.

cbrewster:
@KiChjang probably... In some uses of `create_element`, the intended parent was not immediately obvious to me. Like https://github.com/cbrewster/html5ever/blob/380e38cb078c5ee4eb80be97c18d3d7c7dcc5dd2/html5ever/src/driver.rs#L53-L59

nox:
@bors-servo delegate=cynicaldevil

@cynicaldevil Would you be interested in reviewing this?

html5ever/src/tree_builder/mod.rs
```diff
@@ -1172,17 +1172,17 @@ impl<Handle, Sink> TreeBuilder<Handle, Sink>
...
         let insertion_point = self.appropriate_place_for_insertion(None);
         let (node1, node2) = match insertion_point {
             LastChild(ref p) |
             BeforeSibling(ref p) => (p.clone(), None),
             TableFosterParenting { ref element, ref prev_element } => (element.clone(), Some(prev_element.clone())),
         };
 
+        // Step 7.
+        let qname = QualName::new(None, ns, name);
+        let elem = create_element(&mut self.sink, qname.clone(), attrs.clone(), Some(&node1));
```
cynicaldevil: This might cause problems, because if the `TableFosterParenting` arm is matched in the match block, then either `node1` or `node2` can be the parent. This decision is taken on Servo's side, because we need to access the DOM to figure out which node is the parent.

cbrewster: Hmm true... Would elements be in the same document? Should I send both to `create_element`?

cynicaldevil: These elements are from the [stack of open elements](https://html.spec.whatwg.org/multipage/parsing.html#stack-of-open-elements), where it is possible that two elements may not have the same document associated with them. So you need to pass both elements.

jdm: @cbrewster Does this comment still need to be addressed?

cbrewster: I believe so, its been a while since I worked on these changes. I'm a bit busy at the moment with school, but I'll get back to this in a couple weeks.

xml5ever/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "xml5ever"
-version = "0.12.1"
+version = "0.12.2"
```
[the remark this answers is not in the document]
nox: `0.13`,  `markup5ever` is a public dependency and it got a breaking bump, so the breaking bump needs to be propagated as such.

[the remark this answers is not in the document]
cbrewster: Makes sense, thanks!

cbrewster:
I don't have the bandwidth to finish this at the moment.

PR #313, opened 2017-10-07
Fix WHATWG links

hcpl:


SimonSapin:
Hi! Thanks for this PR. IIRC `/multipage/#foo` is the preferred form: it redirects to the appropriate page and keeps working when something moves to a different page or a page is renamed. Could you change your PR to use that?

hcpl:
I'd be glad to do so, but the whole reason why I put exact links is because `/multipage/#foo` don't work - it just stays at the top of `/multipage`.

SimonSapin:
Humm. I just tried https://html.spec.whatwg.org/multipage/#parsing-html-fragments for example and it worked for me. Do you get some error message in the browser console?

KiChjang:
@hcpl Perhaps you entered it manually when you're already in an HTML spec page? I've been in those situations before and yes, it doesn't work for some strange reason. If you click a spec link like https://html.spec.whatwg.org/multipage/parsing.html#tokenization, it should work though.

SimonSapin:
For what it’s worth, what I tried is opening a new browser tab and copy/pasting the URL into the address bar.

hcpl:
@SimonSapin I found the reason: the anchor redirection requires JavaScript (just read about it [here](https://ffeathers.wordpress.com/2013/09/22/redirecting-anchor-links-on-web-pages/)) while my NoScript filter was on. 

@KiChjang yeah, typing the anchor manually doesn't trigger the script. I think that's because it only tries to redirect at `onload()` and terminates after a timeout.

I guess this is valid enough to prefer `/multipage/#foo` form, even though the JavaScript usage could be documented/warned about on the main page.

SimonSapin:
Thanks.

@bors-servo r+

Issue #315, opened 2017-10-17
Fix warnings in xml5ever


jdm:
```
warning: unused macro definition
   --> target/package/xml5ever-0.11.0/src/tokenizer/mod.rs:609:1
    |
609 | / macro_rules! go_match ( ( $me:ident : $x:expr, $($pats:pat),+ => $($cmds:tt)* ) => (
610 | |     match $x {
611 | |         $($pats)|+ => go!($me: $($cmds)*),
612 | |         _ => (),
613 | |     }
614 | | ));
    | |___^
    |
    = note: #[warn(unused_macros)] on by default

warning: unused macro definition
   --> target/package/xml5ever-0.11.0/src/tokenizer/mod.rs:622:1
    |
622 | / macro_rules! peek ( ($me:expr) => (
623 | |     unwrap_or_return!($me.peek(), false)
624 | | ));
    | |___^

warning: variable does not need to be mutable
   --> target/package/xml5ever-0.11.0/src/tree_builder/mod.rs:359:13
    |
359 |         for mut attr in tag.attrs.iter_mut()
    |             ---^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: #[warn(unused_mut)] on by default
```

milindjain0:
I would like to work on this. I am new to this and help from a mentor for getting started would be appreciated.

Thanks

SimonSapin:
Hi @milindjain0. Are you familiar with GitHub’s pull request workflow?

After cloning this repository, you can run `cargo build` to compile everything. You should see the same warnings as above. Then change the code until you can compile without warnings, and send a pull request.

For the unused macros, double check that they don’t have a `#[macro_export]` attribute above them. This would make them available to users outside the crate, in which case we wouldn’t want to remove them. But I think the warning wouldn’t be shown in that case. You should be able to remove these two macros and have everything still compile.

As to the variable that doesn’t need to be mutable, the warning itself suggests a fix: removing `mut` in `for mut attr in tag.attrs.iter_mut()`.

Feel free to ask any question here on on the `#servo` channel on Mozilla IRC.

milindjain0:
Hi,

I was able to build the code using ./mach build -d and run using ./mach run ...

I am not able to use cargo

Do I need to set environment variables for using Cargo?

Thanks

KiChjang:
[...]

PR #319, opened 2017-10-24
Fix warnings in xml5ever #315

milindjain0:
I have fixed this warning by removing unused marco and removing mut before for loop


jdm:
@bors-servo: r+
Thanks!

Issue #219, opened 2016-10-17
Reduce warning spam


jdm:
Two of the most common warnings I see reported in Servo's automatic reporting system are `stop_parsing not implemented, full speed ahead!` and `FIXME:  /script> not fully implemented`. We should investigate what those mean and remove them if they're not actually useful.

jdm:
https://github.com/servo/html5ever/blob/d4cd4d4598a2e6f00519e1bc8b1f80100a4407c6/src/tree_builder/rules.rs#L795

jdm:
https://github.com/servo/html5ever/blob/d4cd4d4598a2e6f00519e1bc8b1f80100a4407c6/src/tree_builder/actions.rs#L170

Ygg01:
The </script> warning was tied to issues #7 #19.

nox:
`</script>`  has been taken care of. Will look into `stop_parsing`.

nox:
I'm pretty sure `stop_parsing` is useless now, will clean up.

LumaKernel:
Is this already released...? Seems 8 yrs ago, and not sure details, but it's still open...? I'm using 0.27.0, and it seems it's still happening.

PR #322, opened 2017-10-27
Removed warn-log `stop_parsing not implemented, full speed ahead!`

Lakelezz:
As mentioned in #219, `stop_parsing` was assumed to be useless. It still triggers in my project, hence considered removing it.

Hopefully this PR is okay, despite already being assigned to @nox. I simply encountered it a few days ago and decided to make the move, as this issue seems to be there for a while, I preferred to do it myself than bothering anyone for such a small fix : )

Not sure if `"FIXME: </script> not fully implemented"` can be removed as well, if so, just inform me and I will change that as well!


csirkeee:
Is there any reason this has not been merged? It would be useful, those warning messages are annoying.

PR #331, opened 2017-12-07
Add picture local name for picture element

Rakhisharma:
@jdm r?


atouchet:
Is that supposed to be in alphabetical order?

SimonSapin:
Alphabetical order is nice so that there’s no question of where in the file something should go, but it’s not actually important to make the code work. (The strings end up in a static hashset: https://github.com/sfackler/rust-phf)

PR #332, opened 2017-12-07
Update string_cache to 0.7

Eijebong:
Since heapsize isn't used anymore and that string_cache removed its heapsize feature, I removed the feature as a whole meaning that this is a breaking change for h5e and m5e.


SimonSapin:
Thanks!

@bors-servo r+

PR #334, opened 2017-12-07
Fix ordering in local_names.txt

atouchet:


SimonSapin:
Thanks!

@bors-servo r+

PR #349, opened 2018-07-30
Call sink's complete_script method on closing script tag

KiChjang:
Based on https://html.spec.whatwg.org/multipage/parsing.html#scriptEndTag.

### DIFF
--- a/html5ever/src/tree_builder/mod.rs
+++ b/html5ever/src/tree_builder/mod.rs
@@ -340,6 +340,7 @@ impl<Handle, Sink> TreeBuilder<Handle, Sink>
                 }
                 Script(node) => {
                     assert!(more_tokens.is_empty());
+                    self.sink.complete_script(&node);
                     return tokenizer::TokenSinkResult::Script(node);
                 }
                 ToPlaintext => {


nox:
Why?

nox:
https://github.com/servo/html5ever/commit/7a268f3c66ef680323e39b6ade07875b9595479c#diff-940279839ff8d30ec1e7449e12fb9332L774

The `complete_script` was removed because it can't work that way if you want to support `document.write`.

KiChjang:
Hmm... but we definitely need a way to communicate with the Sink that we've read a `</script>` tag, otherwise how else would servo "perform a microtask checkpoint"?

nox:
By doing it when h5e returns Script, just like we do for everything else. See my link.

-- 
Anthony Ramine

nox:
IIRC the only reason `complete_script` is still a method in the API is because we didn't want to go through a breaking bump at the time when @cynicaldevil was working on speculative parsing.

KiChjang:
I'm not sure I follow. The "perform a microtask checkpoint" operation can't be done here in `html5ever`, so it must expose some kind of interface for servo to hook into the parser when it sees a `</script>`.

> By doing it when h5e returns Script, just like we do for everything else. See my link.

Do you mean that this can be done during `create_element` and checking whether the qualified name matches `script`?

nox:
> I'm not sure I follow. The "perform a microtask checkpoint" operation can't be done here in `html5ever`, so it must expose some kind of interface for servo to hook into the parser when it sees a `</script>`.

Yes, and that hook is `html5ever` returning a `TokenSinkResult::Script(node)`.

nox:
> Do you mean that this can be done during `create_element` and checking whether the qualified name matches script?

What does `create_element` have anything to do with the method you wanted to introduce back?

nox:
https://github.com/servo/servo/blob/78e90d6358d2ce2e06862dec9eae1e72bbe2bfab/components/script/dom/servoparser/mod.rs#L470:L473

KiChjang:
Ok, thanks for the pointer! I don't think this PR is needed anymore.

PR #350, opened 2018-09-05
Use log_enabled! to avoid calling to_escaped_string

PluieElectrique:
`to_escaped_string` is somewhat expensive because it allocates twice and iterates over each char individually to escape it. Furthermore, `TreeBuilder::step` produces a lot of `debug_step` messages. So, we use `log_enabled!` to not escape any strings unless debug messages are being printed.


jdm:
@bors-servo r+
Thanks!

PR #353, opened 2018-10-17
Update syn and bump version

Eijebong:


Eijebong:
@dtolnay Mind having a quick look ? Especially the `fork` parts

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
-            |
-            punct!(_) => { |_| None }
-        ) >>
-        punct!(>) >>
-        (
-            Tag {
+impl Parse for Tag {
+    fn parse(input: ParseStream) -> Result<Self> {
+        input.parse::<Token![<]>()?;
+        let closing = input.parse::<Option<Token![/]>>()?;
```
dtolnay: I would write this with the type on the left because `![/]>>()?;` is over the threshold of reasonable adjacent punctuation.

```rust
let closing: Option<Token![/]> = input.parse()?;
```

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
+        let name = if let Ok(i) = input.parse::<syn::Ident>() {
```
dtolnay: I would write this as:

```rust
let name = match input.call(Ident::parse_any)? {
    ref wildcard if wildcard == "_" => None,
    other => Some(other),
};
```

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
+            Some(i)
+        } else {
+            input.parse::<Token![_]>()?;
+            None
+        };
+        input.parse::<Token![>]>()?;
+        Ok(Tag {
                 kind: if closing.is_some() { TagKind::EndTag } else { TagKind::StartTag },
```
dtolnay: This needs to be unindented by one level.

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
-            |
-            many0!(syn!(Tag)) => { |t| LHS::Tags(t) }
-        ) >>
-        (
-            pats
-        )
-    ));
+impl Parse for LHS {
+    fn parse(input: ParseStream) -> Result<Self> {
+        if let Ok(p) = input.fork().parse::<syn::Pat>() {
```
dtolnay: I would write this as:

```rust
if input.peek(Token![<]) {
    let mut tags = Vec::new();
    while !input.peek(Token![=>]) {
        tags.push(input.parse()?);
    }
    Ok(LHS::Tags(tags))
} else {
    let p: Pat = input.parse()?;
    Ok(LHS::Pattern(p))
}
```

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
+        let binding = if input.peek2(Token![@]) {
+            let binding = input.parse::<syn::Ident>()?;
+            input.parse::<Token![@]>()?;
+            Some(binding)
+        } else {
+            None
+        };
+        let lhs = input.parse::<LHS>()?;
+        input.parse::<Token![=>]>()?;
+        let rhs = if let Ok(expr) = input.fork().parse() {
```
dtolnay: This is probably not doing what you intended. This is bleeding over into the next arm and relying on the fact that `{ /* ... */ } < TAG > =>` is not a valid expression -- because `=>` is not a valid beginning of expression. You should parse each arm without parsing the LHS of the next arm as part of the previous arm's expression.

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
+                attrs: vec![],
+                label: None,
+                block,
+            };
+            RHS::Expression(syn::Expr::Block(block))
+        } else {
+            input.parse::<Token![else]>()?;
+            RHS::Else
+        };
+        input.parse::<Option<Token![,]>>()?;
```
dtolnay: This is too lenient and allows commas to be omitted where they should be required. Have the parser require a comma after `else` and after expressions that are not block, unless in the last arm.

html5ever/macros/match_token.rs
```diff
@@ -154,82 +155,89 @@ pub struct Tag {
...
-    named!(parse -> Self, do_parse!(
-        expr: syn!(syn::Expr) >>
-        arms: braces!(many0!(MatchTokenArm::parse)) >> (
-            MatchToken {
-                expr,
-                arms: arms.1
-            }
-        )
-    ));
+impl Parse for MatchToken {
```
dtolnay: Regarding fork -- I don't see anything in the input syntax that would require fork. All of this can be parsed with two tokens of lookahead. Please implement the parsing without using fork to make debugging easier and error messages better.

Eijebong: Way better now. Thanks a lot

SimonSapin:
https://crates.io/crates/html5ever/0.22.5

PR #359, opened 2018-12-04
Expose HtmlSerializer for customizing output

vvilhonen:
This allows 3rd party crates to customize the output without duplicating all the serialization logic. I'm using it myself to pretty print transformed html, where I don't care that the added whitespace changes the meaning of the document.

### DIFF
--- a/html5ever/src/serialize/mod.rs
+++ b/html5ever/src/serialize/mod.rs
@@ -46,14 +46,14 @@ impl Default for SerializeOpts {
 }
 
 #[derive(Default)]
-struct ElemInfo {
+pub struct ElemInfo {
     html_name: Option<LocalName>,
     ignore_children: bool,
     processed_first_child: bool,
 }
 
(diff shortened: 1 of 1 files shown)


jdm:
Could you describe how these changes help your use case? Looking at sample code would be helpful.

vvilhonen:
This is basically my use case https://gist.github.com/vvilhonen/ec87214310578927c41dda7e5ad9269a

html5ever/src/serialize/mod.rs
```diff
@@ -46,14 +46,14 @@ impl Default for SerializeOpts {
 }
 
 #[derive(Default)]
-struct ElemInfo {
+pub struct ElemInfo {
```
jdm: Why does this need to be public?

vvilhonen: A good catch, it actually doesn't and I can't recall what made me turn it public. I restored the original visibility now.

Issue #286, opened 2017-07-03
html5ever is much slower than MyHTML


DemiMarie:
<https://lexborisov.github.io/benchmark-html-persers/> shows html5ever performing very badly against [MyHTML](https://github.com/lexborisov/myhtml), which furthermore has fewer bugs and ~~supports `document.write()`~~ html5ever supports `document.write()` in modern versions (thanks @jdm).

jdm:
html5ever also supports document.write().

DemiMarie:
@jdm That must be an old version I was thinking of, sorry.

SimonSapin:
Sorry, I don’t see the point of this issue. “Fewer bugs” is very non-specific. If you’re aware of bugs in the latest version that aren’t listed in https://github.com/servo/html5ever/issues, it’d be more helpful to file them individually.

As to performance, just stating “very bad” isn’t particularly actionable either. Are you interested in investigating low-hanging fruit to fix performance issues?

nox:
That benchmark doesn't show how `html5ever` was compiled though.

SimonSapin:
Closing for lack of actionable items.

jdm:
The actionable item is attempting to reproduce the results. The benchmark repository looks like it's in good shape for doing that.

Ygg01:
@jdm Perhaps issue #45  #51 might be more appropriate? I did ask about setting up benchmarks in a way that would work best for all parsers, but got no response, so I thought there wasn't any interest.

AFAICT the https://github.com/lexborisov/benchmark-html-persers uses C bindings for html5ever. I thought those were abandoned more than a year ago.

jdm:
https://github.com/swizard0/html2html_lib/blob/master/src/lib.rs is the Rust code that parses the document; the benchmark calls that function from C.

SimonSapin:
The C bindings in this repository were removed a year ago. IIRC they had been broken since we started using string-cache (since string-cache itself doesn’t have a C API), which happened 3 years ago. Also these bindings were only for tokenization.

According to [its `README.md`](https://github.com/lexborisov/benchmark-html-persers/blob/master/README.md), this benchmark uses [swizard0/html2html_lib](https://github.com/swizard0/html2html_lib) which:

* Decodes a UTF-8 byte string
* Parses into an `RcDom` tree
* Optionally serializes to stdout (this is disabled in the benchmark)
* Drops the tree

The tree cannot be accessed from C. (Which might be good enough for a parser.)

[...]

PR #360, opened 2018-12-12
Add criterion.rs to html5ever/xml5ever

### LINKED ISSUE #45: Switch to Criterion.rs for benchmarking
Especially if japaric/criterion.rs#23 is implemented.

Ygg01:
Fixes #45 and is a stepping stone for #286. 

Why is this important for #286?
-----
It allows project to track it's performance, since [criterion.rs](https://github.com/bheisler/criterion.rs) generates benchmark data and tracks changes to performance. 

Possible enhancements
-----
This is first pass at criterion. It's possible the performance could be better tracked, with more generic benchmarks.

Possibly publishing this data, to see how it behaves overall.

More tests for xml5ever, possibly tests that stress namespace rules and parts where xml is weak.


Ygg01:
@jdm Should I separate this PR into two PRs? 

Right now, inside this PR there is a pretty large, overhaul to x5e interface. Mainly to support speculative parsing, but I did it mostly, for interface parity.

html5ever/benches/html5ever.rs
```diff
@@ -0,0 +1,95 @@
...
+        for buf in input.into_iter() {
+            buffer.push_back(buf);
+            let _ = tok.feed(&mut buffer);
+        }
+        let _ = tok.feed(&mut buffer);
+        tok.end();
+    }
+}
+
+fn run_bench(c: &mut Criterion, name: &str, opts: TokenizerOpts) {
```
jdm: It seems odd to take a opts argument and then every caller passes in Default::default.

html5ever/benches/html5ever.rs
```diff
@@ -0,0 +1,95 @@
...
+        // See rust-lang/rust#18045.
+        let sz = std::cmp::min(1024, size - total);
+        input.push(stream.by_ref().take(sz).collect::<String>().to_tendril());
+        total += sz;
+    }
+
+    let mut test_name = String::new();
+    test_name.push_str("tokenizing");
+    test_name.push_str(" ");
+    test_name.push_str(name);
```
jdm: `let test_name = format!("tokenizing {}", name);`

html5ever/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "html5ever"
-version = "0.22.5"
+version = "0.23.0"
```
jdm: I'm pretty sure we don't need a major bump for this change, since we only added dev-dependencies.

html5ever/benches/html5ever.rs
Ygg01: I copied the old code, I thought to leave the option in case someone will need it, in future. I guess YAGNI applies in this case.

html5ever/Cargo.toml
Ygg01: @jdm Ok, I updated to 22.6, is this ok?

html5ever/benches/html5ever.rs
Ygg01: Got it, it's been updated.

html5ever/Cargo.toml
jdm: Yes.

PR #372, opened 2019-04-26
Serialize rcdom iteratively

jdm:
Fixes #371.

PR #374, opened 2019-05-02
Expose public Node constructor.

jdm:
Since #372 there is a private field in Node that prevents constructing them manually. If we want to support that, we need to expose the constructor.

### DIFF
--- a/html5ever/Cargo.toml
+++ b/html5ever/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "html5ever"
-version = "0.22.8"
+version = "0.23.0"
 authors = [ "The html5ever Project Developers" ]
 license = "MIT / Apache-2.0"
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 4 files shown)


jdm:
@SimonSapin Do you have any thoughts about this?

SimonSapin:
It sounds like 0.22.8 was a breaking change and should have been published as 0.23.0 instead, per Semantic Versioning. Now I think we should yank 0.22.8 and publish 0.23.0 with this PR (since `rcdom` becomes less useful if nodes cannot be created except through the parser).

r+ with the version number bump to three crates.

(Or, if we don’t want to have to deal with migrating the ecosystem, let’s find a way to not add fields to `rctree::Node`.)

jdm:
Yeah, I didn't realize that we re-export RcDom in both html5ever and xml5ever, so I only did a semver-major bump for markup5ever. I'll follow your suggestion.

PR #386, opened 2019-07-21
Extract RcDom into its own crate with no support guarantees

### LINKED ISSUE #385: Extract RcDom out of markup5ever
RcDom was never intended to be a production-quality implementation. It's not something that we care about maintaining over time for the purpose, so I want to extract it into its own crate that contains appropriate warnings and maintenance guarantees (read: zero) and remove it as a public API from xml5ever and html5ever.

jdm:
The RcDom implementation was never intended as anything more than a test-only implementation. This work extracts it into a separate crate that contains a README that lays out what sort of support and maintenance guarantees anybody choosing to rely on it can expect (read: none).

Fixes #385.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,5 +2,6 @@
 members = [
     "markup5ever",
     "html5ever",
+    "rcdom",
     "xml5ever"
 ]
--- a/html5ever/Cargo.toml
+++ b/html5ever/Cargo.toml
@@ -37,6 +37,7 @@ serde_json = "1.0"
 rustc-test = "0.3"
 typed-arena = "1.3.0"
 criterion = "0.3"
+markup5ever_rcdom = { version = "0.1", path = "../rcdom" }
 
 [build-dependencies]
 quote = "1"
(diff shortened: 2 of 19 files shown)


jdm:
The first two commits are #384 and can be ignored.

jdm:
We can choose to either publish the markup5ever_rcdom crate or not.
Benefits of publishing:
* makes the html5ever/xml5ever upgrade easier for anybody relying on the library; they just need to add a new dependency and change a couple imports

Downsides of publishing:
* Much easier for people to cargo-cult the dependency from other projects that depend on html5ever and be unaware of the risks associated with using it

Benefits of not publishing:
* forces any users of html5ever to take on the maintenance of some DOM implementation, presumably by forking the existing rcdom code into their own project
* possibly causes someone else to create the one true html5ever DOM crate

Downsides of not publishing:
* user frustration/confusion
* high chance of lots of forked copies of rcdom, making it harder for the ecosystem to deal with new security vulnerabilities that are discovered in the future

nox:
I'm in favour of just killing it and directing people to Simon's Kuchiki crate.

jdm:
Rebased!

PR #390, opened 2019-09-13
Update to 2018 edition and remove all warnings.

jdm:


nox:
r=me but you should probably increase the version too.

jdm:
That can happen the next time it needs to be published.

PR #394, opened 2019-09-30
Adding script tag support to parsing SVGs

### LINKED ISSUE #118: Implement parsing of scripts in SVG

RestitutorOrbis:
Fixes #118 

Currently, this just includes unit tests and a preliminary attempt at a solution. All the unit tests pass except for two tokenizer tests in which I try to parse out the body of the script element (although, I'm not sure if I should even expect that to be parsed out). Are there any other unit tests I should add so that I have a full specification to develop against? In regard to my current solution, it is at this time, just a copy paste of how the HTML parser handles script elements.

### DIFF
--- a/html5ever/src/tokenizer/mod.rs
+++ b/html5ever/src/tokenizer/mod.rs
@@ -1658,6 +1658,110 @@ mod test {
     }
 
     #[test]
+    fn check_script_inside_svg(){
+        let opts = TokenizerOpts {
+            exact_errors: false,
+            discard_bom: true,
+            profile: false,
+            initial_state: None,
+            last_start_tag_name: None,
+        };
+        let vector = vec![
+            StrTendril::from("<svg>"),
+            StrTendril::from("<script>"),
+            StrTendril::from("</script>"),
+            StrTendril::from("</svg>")
+        ];
+        let expected = vec![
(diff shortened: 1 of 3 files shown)


jdm:
Ok, I figured out what's going on here - the LinesMatch tokenizer contains two different vectors, and it uses them inconsistently. The `tokenize` helper is the only code that reads from `lines`, and no code reads from `tokens`. Of course `LinesMatch::finish_str` only pushes to `tokens` and `LinesMatch::push` only pushes to `lines`, so right now there's no single source of truth for all tokens encountered while tokenizing, so the test code will need to be changed in some way - one option would be making sure that `tokens` always contains all of the parsed tokens, and adding a helper function that returns that list instead of `lines` for the tests that care about it.

Some useful tips from my debugging session:
* I added `env_logger` as a dev-dependency to the html5ever Cargo.toml and called `env_logger::init()` from `LinesMatch::new`. This allowed me to run `RUST_LOG=debug cargo test -- --nocapture check_script_body_inside_svg` and see the debug information from the tokenizer when running those tests.
* The code in tree_builder is not invoked by any of the tests in tokenizer. Only the tests in tree_builder will invoke that code.

RestitutorOrbis:
We've decided not to focus on this for now so I'm closing this.

PR #398, opened 2019-10-22
Mark the rcdom crate as not for crates.io

SimonSapin:


nox:
Oh I already set `publish` to `false` in #397.

PR #399, opened 2019-10-23
Allow publishing of markup5ever_rcdom

nox:
Turns out some Android font code uses this with xml5ever in Servo.


SimonSapin:
We ended up replacing that use, in https://github.com/servo/servo/pull/24523.

PR #404, opened 2019-10-30
Don't use serde_json to read the list of named entities for codegen

federicomenaquintero:
This can have the effect of taking out serde_derive from the critical
path of crates that only use markup5ever.  In the case of librsvg,
this reduces build time in release mode by 37 seconds.

Here are two runs of `cargo build -Z timings --release` for librsvg.  Apologies for not attaching the HTML output of `-Z timings`, but github doesn't let one attach HTML files:

* [Unpatched](https://people.gnome.org/~federico/misc/cargo-timing-20191030T151148Z.html) - 149.6s total; markup5ever starts building at around 128s.
* [Patched](https://people.gnome.org/~federico/misc/cargo-timing-20191030T194418Z.html) - 112.5s total; markup5ever starts building at around 51s.


markup5ever/build.rs
```diff
@@ -89,27 +83,16 @@ fn main() {
...
 }
 
-fn named_entities_to_phf(from: &Path, to: &Path) {
-    // A struct matching the entries in entities.json.
-    #[derive(Deserialize, Debug)]
-    struct CharRef {
-        codepoints: Vec<u32>,
-        //characters: String,  // Present in the file but we don't need it
-    }
+include!("entities.rs");
```
jdm: No need to use include! here; we can use `mod entities;` and refer to `entities::NAMED_ENTITIES`.

federicomenaquintero: TIL that one can use mod in build.rs :smiley:  Updated.

federicomenaquintero:
Thanks!

PR #406, opened 2019-12-10
[HTML serializer] Self close void elements

Ahmed-Ali:
At this point starting a [void elements](https://html.spec.whatwg.org/multipage/syntax.html#syntax-start-tag) is treated as any other element (i.e we end up with `<br>` instead of `<br />` and `` instead of ``; which in some use cases isn't backward compatible.

I am not sure if this PR covers it all (n00b here). But at least it demonstrates what I am trying to do.


Ahmed-Ali:
any luck anyone looking into this one?

Ahmed-Ali:
Thanks @jdm for looking into this one. The problem I am trying to solve is that the result of the serialised doc would contain tags like `<br>` instead of `<br />`; in some HTML renderers (epub readers to be specific) this can be troublesome. Some of them will tolerate it, others would treat the whole document as invalid since they expect a closing for the tag (i.e a matching `</br>`)

jdm:
@Ahmed-Ali If those renderers are complaining, then they don't actually want HTML, they want XHTML. You will want to use an XML serializer instead of an HTML one for that.

Ahmed-Ali:
Sorry for my late response.
That definitely makes sense!
Thanks for taking the time to look into it and for the pointer!

PR #408, opened 2019-12-19
Note RcDom deprecation and removal

kornelski:


jdm:
@bors-servo r+
Thanks for noting this!

PR #409, opened 2019-12-30
Don't allocate during rcdom::Node::drop

jameysharp:
Instead of using Vec::extend with user-controlled piles of nodes, which could trigger arbitrarily large heap allocations, use space that was already allocated when constructing the tree: the parent pointers can serve the same purpose as a recursion stack.

Also I think the previous implementation would remove all the children of a node even if there were still other strong references to that node, so this implementation should be more generally correct.

### DIFF
--- a/rcdom/lib.rs
+++ b/rcdom/lib.rs
@@ -125,12 +125,74 @@ impl Node {
     }
 }
 
+fn pop_one_child(node: &Node) -> Option<Handle> {
+    node.children.borrow_mut().pop()
+}
+
 impl Drop for Node {
     fn drop(&mut self) {
-        let mut nodes = mem::replace(&mut *self.children.borrow_mut(), vec![]);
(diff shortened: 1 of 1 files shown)


jameysharp:
I realized that I didn't have to assert that parent pointers were set up correctly, because I could just force them to be set up correctly instead. Now I believe the only way this implementation could panic is if there's a bug in this function itself, and it should give the same results as the default drop handler even if nobody sets parent pointers correctly.

Also I put more thought into explaining the necessary invariants; I think the comments are clearer now.

jdm:
Thanks for the pull request, and apologies for taking so long to provide feedback. Given that rcdom is written with the intention that it's only used by tests in these crates (see https://github.com/servo/html5ever/tree/master/rcdom#markup5ever_rcdom), I don't feel that this complexity is warranted. Arbitrary heap allocations are fine for our purposes.

PR #415, opened 2020-05-11
Added a document parse fuzzer

DavidKorczynski:
This is work that will be coordinated with integration of the [OSS-Fuzz](https://github.com/google/oss-fuzz) infrastructure, which allows continuous fuzzing for the project. 

I had a short talk with @jdm about this. 

Once this has been merged I will set up the project on the OSS-Fuzz side and then we should be up and running :)!

### DIFF
--- a/html5ever/fuzz/fuzz_targets/fuzz_document_parse.rs
+++ b/html5ever/fuzz/fuzz_targets/fuzz_document_parse.rs
@@ -1,9 +1,6 @@
 #![no_main]
 use libfuzzer_sys::fuzz_target;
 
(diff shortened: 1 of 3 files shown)


html5ever/fuzz/fuzz_targets/fuzz_document_parse.rs
```diff
@@ -0,0 +1,38 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+extern crate markup5ever_rcdom as rcdom;
+extern crate html5ever;
```
jdm, line 5: This extern crate isn't necessary, is it?

PR #416, opened 2020-06-06
Drop template contents interatively.

jdm:
This is an extension of the work from https://github.com/servo/html5ever/pull/384.


jdm:
FYI @kornelski

PR #417, opened 2020-06-06
Use a VecDeque for serialization to avoid inefficient array operations.

jdm:
This cuts the time taken for one pathological fuzzing testcase from 9s to 5.5s.


jdm:
Review ping @nox.

nox:
Nice.

@bors-servo r+

PR #422, opened 2020-06-24
Add "line-height"

federicomenaquintero:
This is in CSS3, which is kind of required for SVG2, which I'm slowly getting at in librsvg :smiley:

### DIFF
--- a/markup5ever/local_names.txt
+++ b/markup5ever/local_names.txt
@@ -456,6 +456,7 @@ lighting-color
 limit
 limitingConeAngle
 line
+line-height
 linearGradient
 linebreak
 linethickness


federicomenaquintero:
That was fast, thank you!

SimonSapin:
I don’t find `line-height` in https://svgwg.org/svg2-draft/eltindex.html or https://svgwg.org/svg2-draft/attindex.html, so it doesn’t like it’s an element name or attribute name which is what html5ever uses the `LocalName` type for. Servo uses https://docs.rs/string_cache/0.8.0/string_cache/#with-compile-time-atoms to define a separate `Atom` type for CSS property names, with the names of all supported properties as static atoms. I recommend librsvg does the same, so a PR like this one isn’t needed every time you add support for a new property.

federicomenaquintero:
Oh, `line-height` is indeed a property name, not an attribute name, from https://www.w3.org/TR/CSS22/visudet.html#propdef-line-height

It wasn't clear to me if the names in markup5ever are in fact attribute names that happen to correspond to properties.

Am I doing something wrong here:

* Librsvg stores all attribute names and CSS property names as `QualName`, because the code more or less evolved to that, and because I thought that attributes which have the same name as properties may as well use the same representation for their name.  Should property names live elsewhere / am I wrongly using `QualName` because it seemed like a convenient way to have a namespace/name?
* (Note that librsvg does not implement CSS namespaces yet...)
* If I define my own string cache for CSS property names, won't there be awkward conversions between `QualName`, which expects `type LocalName = Atom<LocalNameStaticSet>`, and my own string cache?  (If the answer above was "`QualName` is not meant for CSS property names", then I guess this question is moot :)

SimonSapin:
CSS properties do not have namespaces. [CSS `@namespace` rules](https://drafts.csswg.org/css-namespaces-3/#declaration) affect selectors, not properties, and is relevant to the namespace of element types and attributes names being matched.

SVG has a number of [presentation attributes](https://svgwg.org/svg2-draft/styling.html#PresentationAttributes) that map to CSS properties of corresponding names, but not all SVG attributes are "presentation" and not all CSS properties relevant to SVG have a corresponding attribute. (Parsing the value of those attributes is also [slightly different](https://svgwg.org/svg2-draft/types.html#presentation-attribute-css-value) from parsing the value of the corresponding property in a stylesheet or `style` attribute context.)

PR #424, opened 2020-07-01
Don't store serialized data in memory when fuzzing.

jdm:
This shouldn't lose much coverage, since we didn't perform any validation of the actual output bytes. By creating a black hole for serialization, we avoid the risk of OOMing when serializing arbitrary fuzzer inputs.

### DIFF
--- a/html5ever/fuzz/fuzz_targets/fuzz_document_parse.rs
+++ b/html5ever/fuzz/fuzz_targets/fuzz_document_parse.rs
@@ -29,7 +29,7 @@ fuzz_target!(|data: &[u8]| {
         return;
     };
 
-    let mut out = Vec::with_capacity(data.len());
+    let mut out = std::io::sink();
     let document: SerializableHandle = dom.document.into();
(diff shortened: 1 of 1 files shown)


html5ever/fuzz/fuzz_targets/fuzz_document_parse.rs
```diff
@@ -29,7 +29,18 @@ fuzz_target!(|data: &[u8]| {
...
+        fn write(&mut self, buf: &[u8]) -> Result<usize, std::io::Error> {
+            self.0 += buf.len();
+            Ok(buf.len())
+        }
+        fn flush(&mut self) -> Result<(), std::io::Error> {
+            Ok(())
+        }
+    }
+
+    let mut out = Writer(0);
```
SimonSapin, line 43: The total length counting appears to be unused. Could this be replaced with `std::io::sink`?

jdm: Sure.

PR #431, opened 2021-01-20
Fix some clippy warnings

svenfoo:
This fixes some Clippy warnings. There are more, and more interesting, warnings in the code base and I would be willing to look into fixing all of them. But first I'd like to know if there's interest in this at all and thus I am opening this pull request with a first set of small changes.


KiChjang:
@bors-servo r+

Thanks!

PR #432, opened 2021-01-21
Fix clippy warnings

svenfoo:
Another wave of changes suggested by Clippy. There are some warnings left, but this is as far as I will get today.

### DIFF
--- a/html5ever/macros/match_token.rs
+++ b/html5ever/macros/match_token.rs
@@ -108,7 +108,6 @@ use std::collections::HashSet;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::path::Path;
-use syn;
 use syn::ext::IdentExt;
 use syn::fold::Fold;
 use syn::parse::{Parse, ParseStream, Result};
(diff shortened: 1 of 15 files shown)


svenfoo:
The build fails because of an outdated Rust version on the Linux build pipeline. I can of course revert the problematic change, it needs at least Rust 1.40. However I would first like to ask what the minimum Rust version is that html5ever is targeting. Is there a good reason to use Rust 1.36 on the CI?

Ygg01:
> Is there a good reason to use Rust 1.36 on the CI?

I think the only valid reason is if servo runs only on 1.36. It's one of few major dependents.

svenfoo:
I have removed two commits that required a version of Rust newer than 1.36.

svenfoo:
Is there a reason why this is not merged? Is there anything you'd like me to change?

svenfoo:
I've added two commits to fix the warnings caused the Travis CI build to fail.

jdm:
The error on nightly is https://github.com/bheisler/TinyTemplate/issues/16.

PR #436, opened 2021-04-15
Remove escape_default workarounds

MichaelMcDonnell:
The `escape_default` method was stabilized in Rust 1.34 [1] and workarounds
are no longer needed.

[1] https://doc.rust-lang.org/std/primitive.str.html#method.escape_default


jdm:
@bors-servo r+
Thanks!

PR #437, opened 2021-04-16
Replace `is_ascii_alnum` with `is_ascii_alphanumeric`

MichaelMcDonnell:
The `is_ascii_alphanumeric` method was added in Rust 1.24 [1] so the
`is_ascii_alnum` function is no longer needed.

[1] https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric


jdm:
@bors-servo r+
Thanks!

PR #440, opened 2021-04-20
Use std `matches` macro

MichaelMcDonnell:
The custom `matches` macro was added before the standard library version was
added. The standard library `matches` macro was added in Rust 1.42 [1]. The
implementations are slightly different but the documentation is very similar.
Additionally, some of the `_tt_as_expr_hack` uses were no longer used with this
change.

[1] https://doc.rust-lang.org/std/macro.matches.html


MichaelMcDonnell:
The build failed because it was building with Rust 1.36.0. Is there a reason that Rust 1.36.0 is required? It is from July 4th 2019.

jdm:
We're moving to 1.49.0 in #469, so this can merge after that one.

PR #444, opened 2021-06-26
Update phf

Morganamilo:
phf 0.8 depends on rand 0.7 This causes rand to be included twice
as most other projects have updated to rand 0.8.

phf 0.9 uses rand 0.8 so update to avoid the duplicate dependency.

### DIFF
--- a/markup5ever/Cargo.toml
+++ b/markup5ever/Cargo.toml
@@ -15,10 +15,10 @@ path = "lib.rs"
 
 [dependencies]
 string_cache = "0.8"
-phf = "0.8"
+phf = "0.9"
 tendril = "0.4"
 log = "0.4"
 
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

jdm:
https://github.com/servo/html5ever/issues/445

JohnTitor:
Note: phf crate bumped MSRV to 1.41 in 0.9.0 so we should also do it here (related comment: https://github.com/servo/html5ever/pull/440#issuecomment-823598182).

Turbo87:
@jdm would it be possible to publish a release including this PR? I've also noticed in the crates.io codebase that this causes quite a bit of duplicate compilation 😅

PR #446, opened 2021-07-02
Migrate CI to GHA

### LINKED ISSUE #445: Move to github actions
Something is wrong with the travis CI integration and homu is no longer starting builds. We should switch to github actions instead.

JohnTitor:
Closes #445
This is almost same as the current Travis confg but has some changes:
- Bumped MSRV to 1.41 as memchr, one of our deps, needs Rust 1.41 on v2.4.0. I think it's fine to bump as it has already affected the downstream crates.
- Runs `cargo doc` only on nightly because I think it doesn't make sense to run it on all the versions (maybe we even want to do with `--no-deps` as it's generally unnecessary).

Since I'm not too familiar with servo's homu, let me know if something is wrong.

CI run on fork: https://github.com/JohnTitor/html5ever/actions/runs/994960129

r? @jdm


jdm:
Thanks for fixing this!

PR #453, opened 2021-10-18
Remove time dependency

xfix:
time 0.1 has known security issues, and it's not really necessary to use time crate anymore for getting precise instants as std::time::Instant::now [has the same implementation as precise_time_ns on popular platforms](https://stackoverflow.com/a/55619763).


jdm:
Addressing the criterion issue in #457.

@bors-servo retry

PR #455, opened 2021-11-15
Reduces log verbosity

sanpii:
My root problem: I use the `html5ever` trough the [scraper crate](https://crates.io/crates/scraper) and I usually display the debug log when I develop, but it’s problematic when I looking for a selector in an entire html page…

I think the trace level is more appropriate for log messages.


jdm:
You can use `RUST_LOG=debug,html5ever=warn` to avoid seeing these.

PR #456, opened 2021-11-16
chore: rust edition 2021 and clippy

Florian-Schoenherr:
a lot, but small changes, see review below for some discussion

### DIFF
--- a/markup5ever/util/buffer_queue.rs
+++ b/markup5ever/util/buffer_queue.rs
@@ -93,10 +93,7 @@ impl BufferQueue {
     /// Look at the next available character without removing it, if the queue is not empty.
     pub fn peek(&self) -> Option<char> {
         debug_assert!(
-            self.buffers
-                .iter()
-                .find(|el| el.len32() == 0)
-                .is_none(),
+            !self.buffers.iter().any(|el| el.len32() == 0),
             "invariant \"all buffers in the queue are non-empty\" failed"
         );
         self.buffers.front().map(|b| b.chars().next().unwrap())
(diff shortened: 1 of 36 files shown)


xml5ever/src/tree_builder/types.rs
```diff
@@ -16,15 +16,15 @@ use crate::tokenizer::{Doctype, Pi, Tag};
...
+    Start,
+    Main,
+    End,
 }
 
 /// A subset/refinement of `tokenizer::XToken`.  Everything else is handled
 /// specially at the beginning of `process_token`.
 #[derive(PartialEq, Eq, Clone, Debug)]
-pub enum Token {
+pub enum TokenWrapper {
```
Florian-Schoenherr, line 27: clippy complained, I think Wrapper makes sense, as XTokens get wrapped by it.
`pub` so might need version bump

xml5ever/src/tree_builder/types.rs
```diff
@@ -16,15 +16,15 @@ use crate::tokenizer::{Doctype, Pi, Tag};
 
 #[derive(PartialEq, Eq, Copy, Clone, Debug)]
 pub enum XmlPhase {
```
Florian-Schoenherr, line 18: clippy complained, I think removing Phase makes sense.
`pub` so might need version bump

markup5ever/util/buffer_queue.rs
```diff
@@ -105,13 +102,13 @@ impl BufferQueue {
     /// Get the next character if one is available, removing it from the queue.
     ///
     /// This function manages the buffers, removing them as they become empty.
-    pub fn next(&mut self) -> Option<char> {
+    pub fn next_char(&mut self) -> Option<char> {
```
Florian-Schoenherr, line 105: clippy complained, thinks it could be confused with `std::iter::Iterator::next`,
also `pub` so if we need a version bump anyways, we could update this as well,
or we can revert

Florian-Schoenherr (commented):
I also just realized a lot of changes are trailing commas and binops, although I haven't changed the `rustfmt.toml`? Haven't overridden it either..

Florian-Schoenherr:
yeah whatever I can't get rustfmt to run

PR #461, opened 2022-01-02
Update phf to 0.10

xfix:


shepmaster:
It would be wonderful if this could be merged and released as it would incidentally resolve https://github.com/integer32llc/rust-playground/issues/780

markup5ever/Cargo.toml
```diff
@@ -15,10 +15,10 @@ path = "lib.rs"
 
 [dependencies]
 string_cache = "0.8"
-phf = "0.9"
+phf = "0.10"
```
shepmaster, line 18: One thing I've done before when the code didn't need to change at all is to have a wide version range:

```diff = "0.10"
- phf = "0.9"
+ phf = { version = ">= 0.9, < 0.11" }
```

xfix: This wouldn't really be correct. Because `phf` has a `phf_codegen` crate it's possible for Cargo to resolve `phf` 0.9 and `phf_codegen` 0.10, which could possibly be bad.

shepmaster: That would be the responsibility of the phf crate to specify which phf_codegen crate it needs. 

If this crate allowed phf {0.8,0.9,0.10} then Cargo could pick any one of those, then would resolve the dependencies for the version it picked.

xfix: `phf_codegen` is a `build-dependency`, and there is nothing stopping Cargo from picking a different version here.

PR #466, opened 2022-02-16
use trace!() in tokenizer, to reduce log spam

alsuren:
html5ever doesn't currently use trace!() anywhere. This patch
downgrades a few of the noisiest debug!() calls to use trace!()
instead.

This seems like a more reasonable log level for a tokenizer.

This makes configuring env_logger to debug other code easier
(RUST_LOG=debug rather than RUST_LOG=html5ever=info,debug), and
makes working with minimal wasm log frameworks like console_log
nicer (the console_log crate doesn't have a way to set different
levels for different log prefixes).

There is a single debug statement in html5ever/src/tokenizer/mod.rs that I've not changed:

        debug!("processing EOF in state {:?}", self.state);

I didn't change this to trace, because I notice this log line anywhere in my logs. If you want me to change this too, then I can.

I'm also happy to close this PR, if you think it's not appropriate.

### DIFF
--- a/html5ever/src/tokenizer/mod.rs
+++ b/html5ever/src/tokenizer/mod.rs
@@ -23,7 +23,7 @@ use self::char_ref::{CharRef, CharRefTokenizer};
 
 use crate::util::str::lower_ascii_letter;
 
-use log::debug;
+use log::{debug, trace};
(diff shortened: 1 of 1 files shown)


jdm (approved):

jdm:
@bors-servo r+
Thanks!

PR #472, opened 2022-04-25
Move code generation into an integration test

djc:
Since the generated code only depends on files in the repository anyway, save the separate compilation for the build script and only generate code in a test. Make it fail if the generated code is different from the stored source code to make CI fail if the generated code is out of date.

If this is deemed useful, we can do a similar thing for the maps generated in markup5ever.


djc:
Ping?

djc:
Rebased.

djc:
@jdm since you seem to be paying attention -- got an opinion on this?

djc:
One other reason I prefer this approach is that the generated code becomes much more legible (for example, via the IDE). I was trying to clean up `extern crate` usage and ended up creating issues in the generated code, which were pretty hard to inspect.

mrobinson:
This is made obsolete by #567, I think.

Issue #464, opened 2022-02-09
`<template />` contents are not serialized


dgp1130:
I was attempting to deserialize an HTML5 document, make some slight modifications, and then reserialize and output it. This document contained `<template />` tags (specifically declarative shadow DOM, but I don't think that matters). I noticed that these `<template />` tags have their content dropped in the output. I made a [minimal reproduction here](https://github.com/dgp1130/html5ever-template). It attempts to deserialize:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Template Element Test</title>
        <meta charset="utf8">
    </head>
    <body>
        <template>
            <div>Hello, World!</div>
        </template>
    </body>
</html>
```

And then reserialize it back out without modification, but instead I get:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Template Element Test</title>
        <meta charset="utf8">
    </head>
    <body>
        <template></template>
    </body>
</html>
```

Note the missing `<div>Hello, World!</div>` from the `<template />` content.

I took a quick look at the source, and I suspect that the serializer [is not aware](https://github.com/servo/html5ever/blob/cfea19f74d922ad049614cdb88a7805bc4ea7e06/html5ever/src/serialize/mod.rs#L215-L235) of `<template />` elements as a special case which don't contain children, but have a special field holding their contents. I'm guessing it will need some `if element.is_template() { write(element.get_template_contents()); }`. I'm not entirely sure where that would go though.

jdm:
Unfortunately this is generally going to be an issue that needs to be addressed within the HTML node serialization algorithm of a crate that depends on html5ever, rather than html5ever itself. rcdom doesn't support serializing templates because we only rely on it for this project's unit tests, and it is [unsupported](https://github.com/servo/html5ever/tree/master/rcdom#markup5ever_rcdom).

PR #479, opened 2022-08-09
Serialize template contents

fabianbieler:
Fixes https://github.com/servo/html5ever/issues/464

I am a bit unsure how best to identify template elements:
- With the `expanded_name` macro as in the pull request
- by comparing `QualName::ns` and `QualName::name` to `ns!(html)` and `local_name!("template")`, respectively
- by checking if `template_contents` is `Some`.

Feedback is appreciated

### DIFF
--- a/rcdom/lib.rs
+++ b/rcdom/lib.rs
@@ -58,6 +58,7 @@ use markup5ever::serialize::{Serialize, Serializer};
 use markup5ever::Attribute;
 use markup5ever::ExpandedName;
 use markup5ever::QualName;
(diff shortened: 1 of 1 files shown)


jdm:
Sorry, rcdom is code that is not published by this project and we only rely on it for our unit tests. https://github.com/servo/servo/blob/cf4f3dd5eb0eb648e071081e96d8272206548f8b/components/script/dom/servoparser/html.rs#L156-L178 is the code that supports correct template serialization in Servo's implementation. Any other user of html5ever will need its own solution in whatever HTML node implementation that it relies upon.

PR #490, opened 2022-12-27
Update phf

mohe2015:
I don't know what the rules for the minimum supported rust version are.


jdm:
There are no specific rules around the MSRV!

@bors-servo r+

PR #507, opened 2023-08-25
Fix parsing of bogus comments after end tags

AcqRel:
This fixes a bug in the tokenizer where the tag name was included in a bogus comment after an appropriate end tag.

For example, this:
```html
<style></style ><!a>
```
is incorrectly parsed as the following:
```html
<style></style><!--stylea-->
```
instead of the expected:
```html
<style></style><!--a-->
```
For this bug to trigger, the end tag needs to be parsed in one of the raw text states (RCDATA, RAWTEXT, or Script data) and have whitespace or a slash after the tag name. I don't know how the contents of the temporary buffer end up inside the comment, but clearing the temporary buffer when exiting the `RawEndTagName` state seems to be enough to fix it.


jdm (approved):
This looks reasonable! Thank you!

PR #508, opened 2023-08-27
Remove redundant FIXME comment

organic1337:
## Background
While searching for potential points in the code to which I could contribute I came across this FIXME comment 
which seemed like an easy-win:

```rust
        // FIXME: linear time search, do we care?
```
(line 511 in `html5ever/src/tokenizer/mod.rs`)

## Fix attempt
I Tries using a HashSet for the attributes names so the search's complexity would be O(1) instead of O(n), as 
can be seen in [this commit](https://github.com/servo/html5ever/commit/df17e5b87ce43f5a3206bf21c4b21ed78276874e)

and once I ran some benchmarks on multiple websites (by [modifying the noop-tokenizer example](https://hatebin.com/bndkyirfnm)) I realized I only made things worst, the Parsing process took a tiny bit longer than before. This is probably because of the overhead within maintaining a HashSet of the attributes names, and because most elements do not have enough attributes for the HashSet to improve performance.

## Conclusion
I think it's a good Idea to remove this comment so no one else will try to optimize this section for no reason.

### DIFF
--- a/html5ever/src/tokenizer/mod.rs
+++ b/html5ever/src/tokenizer/mod.rs
@@ -508,7 +508,6 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
         // Check for a duplicate attribute.
         // FIXME: the spec says we should error as soon as the name is finished.
-        // FIXME: linear time search, do we care?
         let dup = {
             let name = &*self.current_attr_name;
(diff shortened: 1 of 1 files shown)


jdm (approved):

jdm:
Thanks for the analysis!

PR #460, opened 2021-12-10
Update html5lib-tests

untitaker:
Follow-up items:

* More precise errors
* There is still large divergence between some of the states and their names in code vs spec. It's just that there's no testcase for them.

I'm also fuzzing this branch continuously against https://github.com/untitaker/html5gum using `FUZZ_HTML5EVER=1` to find more bugs, but they only cover the tokenizer. Some bugfixes have been made and corresponding regression tests added in a `regression.test` file.

Fixes #459.

PR #516, opened 2024-03-18
html5ever: prepare 0.26.1 release

djc:
Not completely sure if changes from #460 should be considered semver-incompatible? It looks like they add new states but then also the `states` module has these comments:

```
//! This is public for use by the tokenizer tests.  Other library
//! users should not have to care about this.
```

So maybe it doesn't matter?

As far as I can tell other changes are not semver-incompatible.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


mrobinson:
I think this will also need to update the version of markup5ever (as well as the dependency in html5ever/xml5ever) and the version of xml5ever. We need to release new versions of all three crates.

mrobinson (approved):

PR #520, opened 2024-03-21
Check formatting and Clippy lints in CI

djc:
Stacked on the branch for #519 in order to clean up clippy lints in the benchmarks.

### DIFF
--- a/html5ever/src/tokenizer/interface.rs
+++ b/html5ever/src/tokenizer/interface.rs
@@ -19,23 +19,12 @@ pub use self::Token::{EOFToken, NullCharacterToken, ParseError};
 
 /// A `DOCTYPE` token.
 // FIXME: already exists in Servo DOM
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Clone, Debug, Default)]
 pub struct Doctype {
     pub name: Option<StrTendril>,
     pub public_id: Option<StrTendril>,
     pub system_id: Option<StrTendril>,
     pub force_quirks: bool,
-}
-
-impl Doctype {
-    pub fn new() -> Doctype {
-        Doctype {
-            name: None,
-            public_id: None,
-            system_id: None,
-            force_quirks: false,
-        }
-    }
 }
 
 #[derive(PartialEq, Eq, Hash, Copy, Clone, Debug)]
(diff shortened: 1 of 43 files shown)


html5ever/src/tokenizer/interface.rs
```diff
@@ -27,6 +27,12 @@ pub struct Doctype {
...
 }
 
+impl Default for Doctype {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl Doctype {
     pub fn new() -> Doctype {
```
mrobinson, line 37: We should likely remove `Doctype::new()` now and change all references to use the new `Default` implementation.

xml5ever/src/tokenizer/mod.rs
```diff
@@ -1070,9 +1068,8 @@ impl<Sink: TokenSink> XmlTokenizer<Sink> {
             },
             //§ bogus_doctype_state
             XmlState::BogusDoctype => loop {
-                match get_char!(self, input) {
-                    '>' => go!(self: emit_doctype; to Data),
-                    _ => (),
+                if let '>' = get_char!(self, input) {
```
mrobinson, line 1071:
```suggestion
                if '>' == get_char!(self, input) {
```

markup5ever/util/buffer_queue.rs
```diff
@@ -49,6 +49,12 @@ pub struct BufferQueue {
     buffers: VecDeque<StrTendril>,
 }
 
+impl Default for BufferQueue {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
mrobinson, line 57: Same comment here about removing the `new()` function.

mrobinson (changes requested):
Just a couple small comments:

html5ever/src/tokenizer/interface.rs
djc: Good call, done.

xml5ever/src/tokenizer/mod.rs
djc: Heh, yup.

markup5ever/util/buffer_queue.rs
djc: Done.

djc:
(Cleaned up clippy suggestions in rcdom tests/examples after your review.)

mrobinson (approved):

PR #527, opened 2024-04-02
Update README for 0.27

rillian:
Nice to see an update to this crate! Here's a one-liner to update the README example for the new release number.


rillian:
P.S. Thanks for switching the default branch name. :)

PR #528, opened 2024-04-02
Remove rustc-test dependency

### LINKED ISSUE #526: tests fail under rust 1.77.1 to build because of rustc-test
`cargo test` fails to build under rust 1.77.1 because of the `rustc-test` transitive dependency on `rustc-serialize`. Looks like this is already [reported upstream](https://github.com/rust-lang/rust/issues/123156), but I expect the whole crate is no longer necessary and the dependency can removed.

rillian:
The `rustc-test` crate v0.3.1 depends on rustc-serialize v0.3 which is unmaintained and unfortunately no longer compiles under rust 1.77.1.

Remove this crate in favour of a trivial test runner. This drops many nice features of the rustc test runner like parallelization, colour output, per-subtest command line selection support, and output capture on failure. On the other hand, it reduces the maintenance surface and unbreaks the build.

Resolves #526

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,11 +30,6 @@ jobs:
         env:
           RUSTFLAGS: --cfg bench
 
-      - name: Test "rustc-test/capture" feature
-        if: matrix.version == 'nightly'
-        working-directory: rcdom
-        run: cargo test --features "rustc-test/capture"
-
       - name: Cargo test
         if: matrix.version != 'nightly'
         run: cargo test --all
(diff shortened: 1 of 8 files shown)


rillian:
force-pushed a rebase with fixes for ci failures.

jdm (approved):
This looks reasonable. Thanks!

untitaker:
Recommend to move to https://github.com/LukasKalbertodt/libtest-mimic -- currently hacking around the testsuite and the DX is really not good.

PR #506, opened 2023-08-21
Add comments for html5ever/examples

Gangana3:
My first PR here :)

I decided I wanted to contribute to this cool project, and when I started learning about it, I headed over to the examples directory, and it was quite hard for me to understand what I was reading because it lacked some comments.

Someone already opened [an issue](https://github.com/servo/html5ever/issues/451) about it.

**Changes**
- Add comments to the main function in each example with a short description of what the code does.
- Added comments for structs and traits implemented in the examples.
- Move the main function in `arena.rs` example to the end of the file (to match the other examples)

PR #533, opened 2024-04-03
ci: All jobs should pass for CI to pass

mrobinson:
This change makes it so that CI does not pass unless all jobs defined in
the workflow file pass. This is necessary in order to prevent landing
changes that break CI jobs. `success()` and `failure()` which were used
before do not take into account dependent jobs.


rillian (approved):
So the issue is that the merge queue accepted #506 even though the lint job failed? If so this should fix it.

If I understand correctly, the `build_results` job is a way to encode which jobs are required in-tree, so only that job needs to be marked as required in the github settings. It's an extra review burden to check this is set correctly, but as far as I can tell, there's no way to require *all* status checks in a branch protection rule?

mrobinson:
@rillian Thanks for the review. This will also need a review from someone with write access to the repository. I think that for branch protection you have to manually add all jobs that you'd like to gate landing a change. We typically just use this "meta job" approach in Servo. Unfortunately, some of the "meta jobs" are not written correctly, as is the case here.

mukilan (approved):

PR #539, opened 2024-06-05
Only check attributes of current tag for duplicates.

### LINKED ISSUE #538: xml5ever drops xml attributes with matching namespace and local names
Hey there, I've run into this issue recently where only one attribute of a given namespace and local name will remain in the document. Consider the following test
```rs
#[test]
fn test_weird() -> anyhow::Result<()> {
    use rcdom::SerializableHandle;
    use xml5ever::{
        driver::{parse_document, XmlParseOpts},
        serialize::{serialize, SerializeOpts},
        tendril::TendrilSink,
    };

    let src = r##"<svg xmlns="http://www.w3.org/2000/svg" xmlns:x="http://www.w3.org/1999/xlink">
    <defs>
        <g id="mid-line"/>
        <g id="line-plus">
            <use href="#mid-line"/>
            <use href="#plus"/>
        </g>
        <g id="plus"/>
        <g id="line-circle">
            <use x:href="#mid-line"/>
        </g>
    </defs>
    <path d="M0 0" id="a"/>
    <use x:href="#a" x="50" y="50"/>
    <use x:href="#line-plus"/>
    <use x:href="#line-circle"/>
</svg>"##;
    println!("src:\n{src}");

    let dom: rcdom::RcDom =
        parse_document(rcdom::RcDom::default(), XmlParseOpts::default()).one(src);
    let mut sink: std::io::BufWriter<_> = std::io::BufWriter::new(Vec::new());
    serialize(
        &mut sink,
        &std::convert::Into::<SerializableHandle>::into(dom.document),
        SerializeOpts::default(),
    )?;

    let sink: Vec<_> = sink.into_inner()?;
    println!("\noutput:\n{}", String::from_utf8_lossy(&sink));
    Ok(())
}
```

The output for me is as follows
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:x="http://www.w3.org/1999/xlink">
    <defs>
        <g id="mid-line"/>
        <g id="line-plus">
            <use x:href="#mid-line"/>
            <use x:href="#plus"/>
        </g>
        <g id="plus"/>
        <g id="line-circle">
            <use x:href="#mid-line"/>
        </g>
    </defs>
    <path d="M0 0" id="a"/>
    <use x:href="#a" x="50" y="50"/>
    <use x:href="#line-plus"/>
    <use x:href="#line-circle"/>
</svg>

output:
<svg xmlns="http://www.w3.org/2000/svg">
    <defs>
        <g id="mid-line"></g>
        <g id="line-plus">
            <use x:href="#mid-line"></use>
            <use></use>
        </g>
        <g id="plus"></g>
        <g id="line-circle">
            <use></use>
        </g>
    </defs>
    <path d="M0 0" id="a"></path>
    <use x="50" y="50"></use>
    <use></use>
    <use></use>
</svg>
```

As you can see, all the `x:href` attributes, except for the first, seem to be dropped completely.
This doesn't seem like expected behaviour unless I'm missing something.

If needed, this are the versions I'm using
```toml
rcdom = { package = "markup5ever_rcdom", version = "0.3" }
xml5ever = "0.18.0"
```

jdm:
Fixes #538. The current implementation never clears the `present_attrs` field so each unique pair of namespace and local names can only appear once in an XML document. The intent of the implementation is to exclude any attributes that resolve to a duplicate qualified name within an element, so there is no need to store the hashset as a persistent field of the tree builder.


jdm:
Depends on #540 to get rid of unrelated CI errors.

demurgos:
I verified this fix locally, it fixes the issue of missing attributes.

jdm:
> This change makes sense to me. But would it be worth keeping the `HashSet` as a persistent property on the tree and just clearing it at the top of `process_namespaces` as an optimisation?

I prefer the clarity of the current implementation. I'd be willing to change it if there's evidence that it's a performance gap that matters to a user, but I think that could be a followup.

mrobinson:
This caused 29 tests to start passing in Servo! :tada:

PR #542, opened 2024-06-10
Convert BufferQueue to use Interior Mutability

Taym95:
Part of https://github.com/servo/servo/issues/32454

from @jdm [comment](https://github.com/servo/servo/issues/32454#issuecomment-2158242638)

### DIFF
--- a/html5ever/src/tokenizer/mod.rs
+++ b/html5ever/src/tokenizer/mod.rs
@@ -206,7 +206,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
     }
 
     /// Feed an input string into the tokenizer.
-    pub fn feed(&mut self, input: &mut BufferQueue) -> TokenizerResult<Sink::Handle> {
+    pub fn feed(&mut self, input: & BufferQueue) -> TokenizerResult<Sink::Handle> {
         if input.is_empty() {
             return TokenizerResult::Done;
         }
@@ -248,7 +248,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
     //§ preprocessing-the-input-stream
     // Get the next input character, which might be the character
     // 'c' that we already consumed from the buffers.
-    fn get_preprocessed_char(&mut self, mut c: char, input: &mut BufferQueue) -> Option<char> {
+    fn get_preprocessed_char(&mut self, mut c: char, input: & BufferQueue) -> Option<char> {
         if self.ignore_lf {
             self.ignore_lf = false;
             if c == '\n' {
@@ -283,7 +283,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
     //§ tokenization
     // Get the next input character, if one is available.
-    fn get_char(&mut self, input: &mut BufferQueue) -> Option<char> {
+    fn get_char(&mut self, input: & BufferQueue) -> Option<char> {
         if self.reconsume {
             self.reconsume = false;
             Some(self.current_char)
@@ -294,7 +294,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         }
     }
 
-    fn pop_except_from(&mut self, input: &mut BufferQueue, set: SmallCharSet) -> Option<SetResult> {
+    fn pop_except_from(&mut self, input: & BufferQueue, set: SmallCharSet) -> Option<SetResult> {
         // Bail to the slow path for various corner cases.
         // This means that `FromSet` can contain characters not in the set!
         // It shouldn't matter because the fallback `FromSet` case should
@@ -321,7 +321,7 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
     // NB: this doesn't set the current input character.
     fn eat(
         &mut self,
-        input: &mut BufferQueue,
+        input: & BufferQueue,
         pat: &str,
         eq: fn(&u8, &u8) -> bool,
     ) -> Option<bool> {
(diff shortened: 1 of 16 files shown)


jdm (changes requested):
These changes make sense for BufferQueue's internals, but we should also be able to update the usages of `&mut BufferQueue` in the rest of the library. We can then verify that the panic from the testcase in https://github.com/servo/servo/issues/32454#issuecomment-2158235144 should not occur.

Taym95:
> These changes make sense for BufferQueue's internals, but we should also be able to update the usages of `&mut BufferQueue` in the rest of the library. We can then verify that the panic from the testcase in [servo/servo#32454 (comment)](https://github.com/servo/servo/issues/32454#issuecomment-2158235144) should not occur.

I tried to do that but `Iterator` is issue here, `next(&mut sel)`

```
impl Iterator for BufferQueue {
    type Item = char;

    /// Get the next character if one is available, removing it from the queue.
    ///
    /// This function manages the buffers, removing them as they become empty.
    fn next(&mut self) -> Option<char> {
        let (result, now_empty) = match self.buffers.borrow_mut().front_mut() {
            None => (None, false),
            Some(buf) => {
                let c = buf.pop_front_char().expect("empty buffer in queue");
                (Some(c), buf.is_empty())
            },
        };

        if now_empty {
            self.buffers.borrow_mut().pop_front();
        }

        result
    }
}
```
and it is neeed 2 places:

```
    fn eat(
        &mut self,
        input: & BufferQueue,
        pat: &str,
        eq: fn(&u8, &u8) -> bool,
    ) -> Option<bool> {
        if self.ignore_lf {
            self.ignore_lf = false;
            if self.peek(input) == Some('\n') {
                self.discard_char(input);
            }
        }

        input.push_front(mem::take(&mut self.temp_buf));
        match input.eat(pat, eq) {
            None if self.at_eof => Some(false),
            None => {
                self.temp_buf.extend(input);
                None
            },
            Some(matched) => Some(matched),
        }
    }
    ```
    
    ```
        fn eat(&mut self, input: & BufferQueue, pat: &str) -> Option<bool> {
        input.push_front(replace(&mut self.temp_buf, StrTendril::new()));
        match input.eat(pat, u8::eq_ignore_ascii_case) {
            None if self.at_eof => Some(false),
            None => {
                self.temp_buf.extend(input);
                None
            },
            Some(matched) => Some(matched),
        }
    }
    ``

**Comment by jdm** [comment]
Sorry, I missed your last comment when you wrote it. What's the code that actually uses BufferQueue as an iterator?

**Comment by jdm** [comment]
Ah, I see:
```
error[E0277]: `&BufferQueue` is not an iterator
   --> html5ever/src/tokenizer/mod.rs:340:51
    |
340 |                 self.temp_buf.borrow_mut().extend(input);
    |                                            ------ ^^^^^ `&BufferQueue` is not an iterator
    |                                            |
    |                                            required by a bound introduced by this call
    |
    = help: the trait `Iterator` is not implemented for `&BufferQueue`, which is required by `&BufferQueue: IntoIterator`
    = note: `Iterator` is implemented for `&mut BufferQueue`, but not for `&BufferQueue`
    = note: required for `&BufferQueue` to implement `IntoIterator`
note: required by a bound in `extend`
   --> /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/iter/traits/collect.rs:448:5
help: consider dereferencing here
    |
340 |                 self.temp_buf.borrow_mut().extend(*input);
    |                                                   +

error[E0596]: cannot borrow `*input` as mutable, as it is behind a `&` reference
   --> html5ever/src/tokenizer/mod.rs:218:21
    |
218 |                     input.next();
    |                     ^^^^^ `input` is a `&` reference, so the data it refers to cannot be borrowed as mutable
    |
help: consider changing this to be a mutable reference
    |
210 |     pub fn feed(&self, input: &mut BufferQueue) -> TokenizerResult<Sink::Handle> {
    |                                +++
```

**Comment by Taym95** [comment]
> Sorry, I missed your last comment when you wrote it. What's the code that actually uses BufferQueue as an iterator?

Sorry 🙏 , I am not getting notification either, yes the code you shared!

**Comment by jdm** [comment]
I suspect we'll need to use my idea of accepting an argument that can obtain a mutable reference on demand. We might want to try an enum that is either &mut or RefCell and add a method like:
`fn mutate<T, F: FnMut(&mut BufferQueue) -> T>(f: F) -> T` which will call the provider function with the needed mutable reference.

**Comment by jdm** [comment]
Actually, I think the easiest solution is to extract the `next()` method from the `impl Iterator for BufferQueue` so it's just a normal method (and can therefore be `&self` instead of `&mut self`). Then there are only two places in the code that actually try to use it as an iterator, and they can be rewritten.

File: html5ever/src/tokenizer/mod.rs [hunk 1]
```diff
@@ -336,15 +336,17 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         match input.eat(pat, eq) {
             None if self.at_eof => Some(false),
             None => {
-                self.temp_buf.extend(input);
+                while let Some(data) = input.next() {
```
Taym95, line 339: @jdm I extracted `next()` and and reimplemented `extend`

jdm:
Thanks! I'll pull the latest change into my branch and run tests with it.

jdm (approved):

jdm:
Verified that these changes are correct via https://github.com/servo/html5ever/pull/548 and https://github.com/servo/servo/pull/32820, so let's merge this.

PR #548, opened 2024-08-01
Remove mutability from parser/tokenizer APIs.

jdm:
This is incompatible with a re-entrant parsing algorithm which is required by the web platform's document.write API.

### DIFF
--- a/html5ever/examples/noop-tree-builder.rs
+++ b/html5ever/examples/noop-tree-builder.rs
@@ -11,6 +11,7 @@
 extern crate html5ever;
 
 use std::borrow::Cow;
+use std::cell::Cell;
 use std::collections::HashMap;
 use std::io;
 
@@ -19,15 +20,15 @@ use html5ever::tendril::*;
 use html5ever::tree_builder::{ElementFlags, NodeOrText, QuirksMode, TreeSink};
 use html5ever::{Attribute, ExpandedName, QualName};
 
-struct Sink {
-    next_id: usize,
-    names: HashMap<usize, QualName>,
+struct Sink<'a> {
+    next_id: Cell<usize>,
+    names: &'a mut HashMap<usize, QualName>,
 }
 
-impl Sink {
-    fn get_id(&mut self) -> usize {
-        let id = self.next_id;
-        self.next_id += 2;
+impl <'a> Sink<'a> {
+    fn get_id(&self) -> usize {
+        let id = self.next_id.get();
+        self.next_id.set(id + 2);
         id
     }
 }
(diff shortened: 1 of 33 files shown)


jdm:
FYI @Taym95

html5ever/examples/noop-tree-builder.rs
```diff
@@ -60,54 +61,59 @@ impl TreeSink for Sink {
...
     }
 
-    fn elem_name(&self, target: &usize) -> ExpandedName {
-        self.names.get(target).expect("not an element").expanded()
+    fn elem_name(&self, _target: &usize) -> ExpandedName {
+        //XXX(jdm)
+        //let names = self.names.borrow();
+        //Ref::map(names, |names| names.get(target).expect("not an element").expanded())
+        //self.names.get(target).expect("not an element").expanded()
+        todo!()
```
jdm, line 69: I still need to figure out what to do about this code.

html5ever/examples/print-tree-actions.rs
```diff
@@ -41,69 +42,71 @@ impl TreeSink for Sink {
...
     fn same_node(&self, x: &usize, y: &usize) -> bool {
         x == y
     }
 
-    fn elem_name(&self, target: &usize) -> ExpandedName {
-        self.names.get(target).expect("not an element").expanded()
+    fn elem_name(&self, _target: &usize) -> ExpandedName {
+        //XXX(jdm)
+        //self.names.borrow().get(target).cloned().expect("not an element").expanded()
+        todo!()
```
jdm, line 73: Same here.

html5ever/src/macros.rs
```diff
@@ -9,9 +9,9 @@
 
 macro_rules! unwrap_or_else {
     ($opt:expr, $else_block:block) => {
-        match $opt {
-            None => $else_block,
-            Some(x) => x,
+        {
+            let Some(x) = $opt else { $else_block };
+            x
```
jdm, line 14: This change is necessary to minimize the scope of refcell borrows that are passed as arguments to the macro.

jdm:
> > FYI @Taym95
> 
> Thanks, Should I close the other PR? lets me knew is you need help here in clean up.

We can still merge the other PR, but we should remove all uses of `&mut BufferQueue` from the codebase in it (and uses like `&mut input`).

jdm:
Ok, the code's cleaned up and commits are split into:
1. mechanical changes - lots of removing `&mut self` from methods and adding Cell/RefCell
2. new APIs needed by https://github.com/servo/servo/pull/32820
3. functional changes - the TreeSink API for obtaining an element's name relies on borrowed data in ways that do not play nicely with naive usage of RefCell to store the element name data ([ExpandedName docs)](https://docs.rs/markup5ever/latest/markup5ever/interface/struct.ExpandedName.html)
4-7. getting CI to pass

jdm:
@Taym95 Would you like to review this?

Taym95 (approved):
I've reviewed all the changes, and everything looks good to me!

mrobinson (approved):

PR #551, opened 2024-08-30
Don't ignore warnings in `html5ever/src/tree_builder`

Wuelle:
Currently, `html5ever/src/tree_builder/mod.rs` contains this:
https://github.com/servo/html5ever/blob/e6f0018d50b358240392a44851972b66c69deeb5/html5ever/src/tree_builder/mod.rs#L10

Presumably, this is done to ignore all the warnings caused by the `rules.rs` file generated at build time (Which is simply included in `mod.rs`).

However, this masks *many* warnings from the actual code itself. This PR instead puts the generated code into its own module, ignores warnings *only* for this specific module and then fixes (almost) all the warnings in the rest of the code.
```rust
#[allow(warnings)]
mod autogenerated {
    include!(concat!(env!("OUT_DIR"), "/rules.rs"));
}
```

There are two warnings that are not fixed by this PR:
* all the `Token` variants ending with `Token` themselves (`Token::TagToken`, `Token::EOFToken`...). I didn't fix this because these variants are used directly without the enum name (`TagToken`, `EOFToken`...). Not a fan, but outside of the scope of this PR
* Some enum variants are never constructed, I've ignored this in https://github.com/Wuelle/html5ever/commit/86dcdeb992bfc1d72cfb0448abd3fda04a94ed0b since I'm not sure if these can actually be removed

I've made individual commits for each "kind" of lint being addressed - I can squash these together if preferred.


mrobinson (approved):
Nice!

PR #553, opened 2024-09-06
Make return type of `TreeSink::elem_name` an associated type.

### LINKED ISSUE #552: Problem using new immutable API with Slab-based DOM tree
I am using a `slab::Slab` to store nodes for my DOM implementation. The new immutable API in html5ever 0.28 has required me to wrap my node tree in a `RefCell` so that I can mutate it from html5ever's trait methods. However, html5ever's `ExpandedName<'a>` type also requires me to return a (plain) reference into my tree, which is not possible if my tree is wrapped in a `RefCell`.

I believe the best fix for this (assuming that the new interior mutability-based API is indeed necessary - it seems awkward!) is to make `ExpandedName` an associated type on the `TreeSink` trait which implements `AsRef<ExpandedName>` or alternatively has methods to access the fields of ExpandedName individually. This would allow implementers of `TreeSink` to choose one of the following strategies:

- Use the existing ExpandedName type if their tree supports it.
- Use a variant of ExpandedName that uses a RefCell `Ref` (or `MutexGuard`, etc)
- Return cloned/owned names

I would be happy to put up a PR for this if this seems like a reasonable solution. ~I have not yet completed my migration, so I am unsure if I will run into further problems.~ I have now migrated the rest of the code. This seems to be the only thing I couldn't solve with a `RefCell`.

nicoburns:
Fixes #552

This allows this type to be owned or contain data protected by a RefCell or Mutex. It can also simply be `ExpandedName` - the type which was previously the return type of `TreeSink::elem_name`.

I have not yet had a chance to test this, but I believe this ought to fix my issue.

### DIFF
--- a/markup5ever/interface/mod.rs
+++ b/markup5ever/interface/mod.rs
@@ -8,12 +8,13 @@
 // except according to those terms.
 //! Types for tag and attribute names, and tree-builder functionality.
 
+use std::cell::Ref;
 use std::fmt;
 use tendril::StrTendril;
 
 pub use self::tree_builder::{create_element, AppendNode, AppendText, ElementFlags, NodeOrText};
+pub use self::tree_builder::{ElemName, NextParserState, Tracer, TreeSink};
 pub use self::tree_builder::{LimitedQuirks, NoQuirks, Quirks, QuirksMode};
-pub use self::tree_builder::{NextParserState, Tracer, TreeSink};
 use super::{LocalName, Namespace, Prefix};
 
 /// An [expanded name], containing the tag and the namespace.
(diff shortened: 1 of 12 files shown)


nicoburns:
An unrelated clippy lint was failing. I've added an allow, although it seems to be a private item so we could update the variant names.

nicoburns:
Hmm... I don't think this actually works with `AsRef` (seems impossible to actually implement the trait sensibly). I think we'll need a custom trait.

markup5ever/interface/mod.rs
```diff
@@ -35,6 +60,66 @@ impl<'a> fmt::Debug for ExpandedName<'a> {
...
+//     fn as_ref(&self) -> &ExpandedName<'a> {
+//         self
+//     }
+// }
+
+// /// A wrapper around an `ExpandedName` which allows it to be borrowed from a RefCell
+// pub struct ExpandedNameRef<'a>(pub Ref<'a, ExpandedName<'a>>);
+
+// impl<'a> AsRef<ExpandedName<'a>> for ExpandedNameRef<'a> {
+//     fn as_ref(&self) -> &ExpandedName<'a> {
```
mrobinson, line 73: Out of curiosity, why is this code commented?

nicoburns: This was a dead-end experiment (it ended up being impossible to actually implement AsRef in the relevant circumstances). The new approach with a custom trait is working for me. I'll go through and clean up later.

markup5ever/interface/tree_builder.rs
```diff
@@ -113,6 +128,11 @@ pub trait TreeSink {
     /// [rust-lang/rust#29661](https://github.com/rust-lang/rust/issues/29661)
     type Output;
 
+    //
```
jdm, line 131:
```suggestion
    /// A type that encapsulates an element's local name and namespace.
```

jdm (approved):
This is a great improvement. Thanks!

markup5ever/interface/mod.rs
nicoburns: Done!

markup5ever/interface/tree_builder.rs
nicoburns: I've pushed a similar comment :)

PR #560, opened 2024-10-30
Relax selector parsing

untitaker:
Neither the proposal nor the test changes are merged yet. Also, the
tests are still failing.

https://github.com/html5lib/html5lib-tests/pull/178
https://github.com/whatwg/html/pull/10557

### DIFF
--- a/html5ever/src/tree_builder/mod.rs
+++ b/html5ever/src/tree_builder/mod.rs
@@ -1146,6 +1146,7 @@ where
         n
     }
 
+    /// Pop element until an element with the given name has been popped.
     fn pop_until_named(&self, name: LocalName) -> usize {
         self.pop_until(|p| *p.ns == ns!(html) && *p.local == name)
     }
@@ -1231,16 +1232,6 @@ where
                 _ => continue,
             };
             match *name {
-                local_name!("select") => {
-                    for ancestor in self.open_elems.borrow()[0..i].iter().rev() {
-                        if self.html_elem_named(ancestor, local_name!("template")) {
-                            return InSelect;
-                        } else if self.html_elem_named(ancestor, local_name!("table")) {
-                            return InSelectInTable;
-                        }
-                    }
-                    return InSelect;
-                },
                 local_name!("td") | local_name!("th") => {
                     if !last {
                         return InCell;
--- a/html5ever/src/tree_builder/rules.rs
+++ b/html5ever/src/tree_builder/rules.rs
@@ -624,6 +624,13 @@ where
 
                 tag @ <hr> => {
                     self.close_p_element_in_button_scope();
+                    if self.in_scope_named(default_scope, local_name!("select")) {
+                        self.generate_implied_end_except(local_name!("optgroup"));
+                        if self.in_scope_named(default_scope, local_name!("option")) {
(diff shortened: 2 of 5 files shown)


nicoburns:
@untitaker I'm not sure if you still have time to work on this, but spec proposal has now been merged, and we'd love to get `html5ever` updated to match. There are conflicts here because we eliminated the proc macro from `rules.rs` (see https://github.com/servo/html5ever/pull/640), but this is a really just a mechanical/syntactical change that should be fairly easy to adapt to.

I also wanted to notify you that this crate has seen a [revival in activity and maintainence](https://github.com/servo/html5ever/graphs/code-frequency) in the last few months, and to invite you to collaborate with us if you are interested in doing so.

untitaker:
i'm stumped on how to implement this part of the spec which seems to be part of handling `</option>` in in-body insertion mode: https://html.spec.whatwg.org/multipage/form-elements.html#maybe-clone-an-option-into-selectedcontent

to me this implies that the `TreeSink` trait needs methods for traversing the ancestors of the tree. Is that right? Then this is a breaking change.

nicoburns:
> to me this implies that the TreeSink trait needs methods for traversing the ancestors of the tree. Is that right?

I haven't looked into this deeply, but we already keep track of the currently open `<form>` element so that we can associate elements with it (https://github.com/servo/html5ever/blob/main/html5ever/src/tree_builder/mod.rs#L120). Could a similar approach be used for `<select>`?

> ...is a breaking change

If it is a breaking change then that's probably not a big deal: html5ever makes breaking changes [quite frequently](https://lib.rs/crates/html5ever/versions)

untitaker:
Progress update. Indeed the conflict resolution was trivial, but since I opened the PR, the corresponding spec PR has significantly expanded in scope. So it's not all easy.

* The recently added tests for `<selectedcontent>` contain this input:

  ```
  <select><button><selectedcontent></button><option>X
  ```

  And expect `X` to be duplicated into `<selectedcontent>`. However, according to the spec, the routine responsible for that ("maybe clone option...") is only called from within the transition for `</option>` end tags. so now I'm confused where that `</option>` token is coming from? it's not coming from the tokenizer, and from what i can tell nothing in the spec says the treebuilder has to fabricate it. is it dynamically inserted post-tree-building?
  
  EDIT: https://github.com/whatwg/html/issues/11653
  
* the "selectedness" of a `<option>` element influences what gets cloned into `<selectedcontent>`. however, "selectedness" is _not part of tree building_. the selectedness attribute is set to true based on the [selectedness algorithm](https://html.spec.whatwg.org/#selectedness-setting-algorithm). The only way I can see it being part of tree building is if we consider the `selected` attribute from the original input, but in the testcase there is none. The testcase just assumes we're doing way more than treebuilding.

Am I missing something?

filed https://github.com/html5lib/html5lib-tests/issues/180

untitaker:
in order to move forward with updating html5lib-tests i ignored the remaining two tests. they are testing full DOM behavior which html5ever does not care about.

untitaker:
@nicoburns i think this can be reviewed/merged

jdm (approved):
This is great work. Thank you for doing it, and sorry for sitting on it for so long without feedback!

PR #567, opened 2025-01-08
Replace manual code generation with procedural macro

simonwuelker:
This makes LSPs like rust-analyzer pick up the `rules.rs` file, making it much more pleasant to edit.

It also completely removes the `build.rs`, which has many benefits (less mental overhead, faster compile times, easier integration with non-cargo build systems, ...)


match_token/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "match_token"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+syn = "2"
```
Mrmaxmeier, line 7: This used to be `syn = { version = "2", features = ["extra-traits", "full", "fold"] }`, and it seems like (some?) these extra features are (still?) necessary.

I'm seeing this error when building html5ever from git:
```
error[E0433]: failed to resolve: could not find `Pat` in `syn`
   --> /projects/html5ever/repo/match_token/src/lib.rs:144:37
    |
144 |             let p = input.call(syn::Pat::parse_single)?;
    |                                     ^^^
    |                                     |
    |                                     could not find `Pat` in `syn`
    |                                     help: a struct with a similar name exists: `Path`
    |
note: found an item that was configured out
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/syn-2.0.96/src/lib.rs:484:15
    |
484 |     FieldPat, Pat, PatConst, PatIdent, PatLit, PatMacro, PatOr, PatParen, PatPath, PatRange,
    |               ^^^
note: the item is gated behind the `full` feature
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/syn-2.0.96/src/lib.rs:481:7
    |
481 | #[cfg(feature = "full")]
    |       ^^^^^^^^^^^^^^^^

error[E0412]: cannot find type `Pat` in crate `syn`
```

Haven't looked into why this error only shows up in my situation though :shrug:

simonwuelker: These errors don't show up when building `html5ever` because the `full` feature is enabled by some other dependency  (probably `serde-derive`). `cargo b -p match_token` fails - I'll open a PR to fix that.

Mrmaxmeier: Ah right, looks like some innocent dependency dropped the `full` feature in a minor version bump. :slightly_smiling_face:

PR #570, opened 2025-01-21
Enable "full" feature for syn

simonwuelker:
Fixes the build issue brought up in https://github.com/servo/html5ever/pull/567#discussion_r1924372675

PR #571, opened 2025-01-21
Fix fmt and clippy errors

simonwuelker:


mrobinson:
Hrm. Maybe this needs to be landed together  with #570? Otherwise, we could disable the merge queue checks just for one of these and then immediately re-enable them.

mrobinson (approved):

simonwuelker:
I'll combine them together

Issue #573, opened 2025-02-12
markup5ever non-deterministic generated.rs


bmwiedemann:
Originally filed in https://github.com/lycheeverse/lychee/issues/1632 - check out the details there.

Please ensure that markup5ever's `generated.rs` is created in a deterministic way, to allow for reproducible builds.

Ygg01:
This will be a bit more tricky than advertised.

What seems to be the source of non-determinism is the `phf` turning https://github.com/servo/html5ever/blob/main/markup5ever/entities.rs into an perfect hash map. 
https://github.com/servo/html5ever/blob/0e4e125b8672abad4ff14d1e0043d3be9305af6a/markup5ever/build.rs#L36


The problem I see is, I don't know where the non-determinism is coming from. It could be as simple as changing `HashMap` into `TreeMap`:
https://github.com/servo/html5ever/blob/0e4e125b8672abad4ff14d1e0043d3be9305af6a/markup5ever/build.rs#L82-L88
 or as hard as getting `phf` or `string_cache_codegen` to emit stable perfect hash map.

bmwiedemann:
Rust HashMaps have random iteration order. https://github.com/osa1/tiny/pull/438 solved that with an indexmap. Though it would be nice, if rust had some builtin way for that.

nicoburns:
> Rust HashMaps have random iteration order

Yeah, there's a strong chance that it's that `HashMap` causing this non-determinism.

> though it would be nice, if rust had some builtin way for that.

There is `std::collectionsBTreeMap` would likely work well in this case.

Ygg01:
> that `HashMap` causing this non-determinism.

The big thing is how do we repro this non-determinism?

bmwiedemann:
In openSUSE, I build our packages with
```
osc checkout openSUSE:Factory/lowfi && cd $_
for N in 1 2 ; do
  osc build --debuginfo --vm-type=kvm --clean --noservice --keep-pkgs=RPMS.$N --release=1.1 standard
done
```
but it will probably be easier for you with some `cargo build` of https://github.com/talwat/lowfi or slower https://github.com/lycheeverse/lychee

bmwiedemann:
So when can we expect a new version with [servo/string-cache#290](https://github.com/servo/string-cache/pull/290) integrated? With `nushell` the third affected package is just coming to openSUSE:Factory

Ygg01:
I don't know the process, I can ask around Zulip for deets.

PR #574, opened 2025-02-16
Use `BTreeMap` rather than `HashMap` in markup5ever build script

nicoburns:
Speculative fix for #573

@bmwiedemann could you test and see whether this PR fixes your issue?

### DIFF
--- a/markup5ever/build.rs
+++ b/markup5ever/build.rs
@@ -10,7 +10,7 @@
 extern crate phf_codegen;
 extern crate string_cache_codegen;
 
-use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


bmwiedemann:
Unfortunately, it did not help.

Ygg01:
I'm about 80% certain this + https://github.com/servo/string-cache/pull/290 will close this issue.

`string_cache_codegen` was behaving non-deterministically.

bmwiedemann:
In my tests, this change was not needed. Only the string-chache one.

nicoburns:
I think we might as well have this. It can't hurt.

PR #578, opened 2025-02-21
Update declarative shadow dom

longvatrong111:
Servo PR:  https://github.com/servo/servo/pull/34964

cc: @xiaochengh, @simonwuelker

PR #581, opened 2025-03-11
Publish html5ever 0.29.2 and markup5ever 0.15.0

longvatrong111:
https://github.com/servo/html5ever/pull/578 made a breaking change to markup5ever: [any change to trait item signatures](https://doc.rust-lang.org/cargo/reference/semver.html#trait-item-signature)

### DIFF
--- a/html5ever/Cargo.toml
+++ b/html5ever/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "html5ever"
-version = "0.29.1"
+version = "0.29.2"
 authors = [ "The html5ever Project Developers" ]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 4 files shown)


jdm (approved):

longvatrong111:
xml5ever still depends on markup5ever@0.14.1. Seem like I need to bump version of xml5ever too?

jdm:
Ah, yes.

nicoburns:
@longvatrong111 @jdm I believe this should have been a minor version bump for `html5ever` and `xml5ever` too. Those crates both re-export the `TreeSink` trait that had breaking changes, and thus I believe the latest release of those crates also contained semver-breaking changes.

We should probably republish as `0.30` and `0.21`.

jdm:
Oof, you're right. I'll yank those versions.

Issue #580, opened 2025-03-09
example print-rcdom parses <pre> contents


qknight:
The example `print-rcdom` shows that it parses the contents of the `<pre>`. In HTML it is usually only relevant for browsers using the given CSS for styling but not as a general concept.

My use-case is that I want to have arbitrary data in the `<pre><code>` section, even invalid html and I don't want any interpretation of it. See https://lastlog.de/blog/vscodium_and_nix_develop.html where I have bash output inside these boxes.

So my question is:

**How to parse `<pre>` content as NodeData::Text instead of _yet_ another bunch of HTML tags?**

The example below shows that all the html code is actually parsed.

```bash
[nixos@nixos:~/html5ever/rcdom/examples]$ cargo run --example print-rcdom
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `/home/nixos/html5ever/target/debug/examples/print-rcdom`
<div><p> test </p><pre><code>
<p>foo1</p>
  <p>foo2</p><p>foo3</p>
  3</code></pre></div>
#Document
    <html>
        <head>
        <body>
            <div>
                <p>
                    #text:  test 
                <pre>
                    <code>
                        #text: \n
                        <p>
                            #text: foo1
                        #text: \n  
                        <p>
                            #text: foo2
                        <p>
                            #text: foo3
                        #text: \n  3
            #text: \n

Parse errors:
    Unexpected token
```

qknight:
The html5ever implementation for <pre> is alright and my problem was caused by the post-processing of the let dom = parse_document(RcDom::default(), opts).one(input); later on.

PR #582, opened 2025-03-11
Attempt of getting <pre> to not parse inner contents similar to <script>

qknight:
This branch is used to implement the fix required to the issue: https://github.com/servo/html5ever/issues/580

# Motivation

With the current implementation the parser will evaluate arbitraty html tags inside a `<pre>...</pre>` and with this patch, `<pre>` will behave more like `<script>`.

This behaviour should be optional as sometimes it also makes sense to parse tags inside a `<pre>`, for instance for styling but most often the content inside a `<pre>` should be pretty much ignored and copied 1:1 from the source document into the generated output document and not reformatted (removing spaces, newlines or tabs) or should the parsed content have any influence on the overal consistenty of the document. 

That said: 

* `<html><pre></html>test   foo</pre></html>` should not be fixed into 
* `<html><pre>test foo</pre></html>`

# Status

branch: **servo_issue_580** with hash: **2094a85f5260b23397c4a89fbf50dcba883219e6**

This evaluates: 

`<hello>XML</hello><pre>\n<bad> </bad>text-in  pre</pre><p>asdf</p><script>script</html> magic string</script>`

into

`<html><head></head><body><hello>XML</hello><pre>\n&lt;bad&gt; &lt;/bad&gt;text-in  pre</pre><p>asdf</p><script>script</html> magic string</script></body></html>`

This shows that the content inside the `<pre>...</pre>` is grabbed and not parsed already. Yet the result should be no HTML escaped string but rather a 1:1 copy of the original tags.

This can be evaluated by running:

`clear && cargo run --example html2html`

# Todo

* [x] Figure out why: `process_to_completion` is called for `<script>` but not for `<pre>`
* [x] Implement an option to the parser to include parsing of `<pre>...</pre>` content or not
* [ ] Write the PreData as String and not HTML escaped.
* [x] Write a bunch of tests so make sure it works

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target
 .vscode
 Cargo.lock
 *.racertmp
+.direnv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,91 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "id": "flake-utils",
+        "type": "indirect"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 0,
+        "narHash": "sha256-edm8WG19kWozJ/GqyYx2VjW99EdhjKwbY3ZwdlPAAlo=",
+        "path": "/nix/store/b309kyd5hrvrrf5z5xbqglmb1hhm8n6p-source",
+        "type": "path"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "nixpkgs_2": {
+      "locked": {
+        "lastModified": 1736320768,
+        "narHash": "sha256-nIYdTAiKIGnFNugbomgBJR+Xv5F1ZQU+HfaBqJKroC0=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "4bc9c909d9ac828a039f288cf872d16d38185db8",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixpkgs-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs",
+        "rust-overlay": "rust-overlay"
+      }
+    },
+    "rust-overlay": {
+      "inputs": {
+        "nixpkgs": "nixpkgs_2"
+      },
+      "locked": {
+        "lastModified": 1737685583,
+        "narHash": "sha256-p+NVABRpGi+pT+xxf9HcLcFVxG6L+vEEy+NwzB9T0f8=",
+        "owner": "oxalica",
+        "repo": "rust-overlay",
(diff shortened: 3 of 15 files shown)


jdm:
Is this behavior specified in the HTML parsing specification?

qknight:
@jdm your question is hard to answer!

# html standard related to `<pre>`

* https://w3c.github.io/html-reference/pre.html
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre

i like the grok summary i created https://x.com/i/grok/share/AI7crMuXH2BoIAxC57P9v8VIg but it does not have sources.

my new understanding is now:

* everything in `<pre>...</pre>` needs to have a fixed layout, no changes on spaces, tabs or newlines
* the parser 'can' parse tags but must not do any 'fixes' if incorrect

something along these lines. i try to figure out how virtual-dom does it.

# virtual-dom (works)

i write this technical blog at https://lastlog.de/blog/libnix_volth's_work.html and i'm using pandoc to generate `<pre><code>` sections and when i serialize and deserialize the html document using https://github.com/Matt-Esch/virtual-dom it just works correctly.

the motivation to move away from this is the usage of rust compiled to WASM. i always wanted to make modifications to the way 'new virtual-dom patches are applied' with visual cues which i can't do with virtual-dom.

# rphtml (fails)

first i tried to replace virtual-dom with _rphtml_. but i discovered problems with rphtml: https://github.com/fefit/rphtml/issues/4 
i tried to fix them but the code is very hard to read and after a few days of hacking i gave up.

notable mention: the issue of rphtml was very hard to track down as it works 'half' of the time where text nodes in combination to tags sometimes yield correct html documents after `doc.render(...)`

# html5ever

the `<pre>` handling in html5ever also breaks the code generated by pandoc after a serialize/deserialize run but for a slightly different reason. it removes all the newlines from the `<pre>... \n <span>...</span> \n ... </pre>`.

the first attempt in fixing it using a string for all the `<pre>...</pre>` content might work but still might not implement the html parsing standard correctly.

qknight:
@jdm I checked the tests in html-serializer.rs and they seem correct!

```
test!(pre_lf_0, "<pre>foo bar</pre>");
test!(pre_lf_1, "<pre>\nfoo bar</pre>", "<pre>foo bar</pre>");
test!(pre_lf_2, "<pre>\n\nfoo bar</pre>", "<pre>\nfoo bar</pre>");
test!(pre_lf_3, "<pre>\n  <p>adf</p>\nfoo\n\tbar</pre>", "<pre>  <p>adf</p>\nfoo\n\tbar</pre>");
```

I think that html5ever handles it correctly and in sauron, in order to do DOM / vDOM diff/patches it is translated into a different node abstraction. I've added a check on the tag name for "pre" and execute this code instead:

```rust
fn process_handle<MSG>(node: &Handle) -> Result<Option<Node<MSG>>, ParseError> {
    let children: Vec<Node<MSG>> = node
        .children
        .borrow()
        .iter()
        .filter_map(|child| process_handle(child).ok().flatten())
        .collect();

    match &node.data {
        NodeData::Document => {
            let child_nodes_len = children.len();
            match child_nodes_len {
                0 => Ok(Some(node_list([]))),
                1 => Ok(Some(children.into_iter().next().unwrap())),
                _ => Ok(Some(node_list(children))),
            }
        }
        NodeData::Text { contents } => {
            let content = contents.borrow().to_string();
            Ok(Some(text(content)))
        }
        NodeData::Element { name, attrs, .. } => {
            let tag_name = name.local.to_string();
            
            if tag_name == "pre".to_string() {
                //println!("tag_name: {}", tag_name);
                let mut buffer: Vec::<u8> = vec![];
                let document: SerializableHandle = node.clone().into();
            
                serialize(&mut buffer, &document, Default::default()).expect("serialization failed");
                let writer_string = String::from_utf8(buffer).expect("Could not write buffer as string");
                println!("--- {} ---", writer_string);
                let content: String = format!("<pre>{}</pre>", writer_string);
                Ok(Some(text(content)))
            }
            else {
```

In words: When a `<pre>` tag occures, I use `serialize` / `SerializableHandle` to convert it into a correctly formated String and insert it as NodeData::Text. Not yet sure this does exactly what I want but so far it is looking good. I'll close this ticket shortly if that is the case.

qknight:
The html5ever implementation for `<pre>` is alright and my problem was caused by the post-processing of the `let dom = parse_document(RcDom::default(), opts).one(input);` later on.

PR #590, opened 2025-03-31
Handle decoding of input in `html5ever`

simonwuelker:
These changes are an attempt to allow users of `html5ever` to respect the encodings specified with <meta charset="...">` tags in a spec-compliant way.

The major change is that the https://html.spec.whatwg.org/#input-stream now lives in the html5ever crates. In effect,  the new API surface exposes a "pull" instead of a "push" interface.

The entry point to the new API is a `DecodingParser`, which wraps either a HTML or an XML parser.
After providing some amount of byte input to a `DecodingParser`, the caller can call `DecodingParser::parse`, which returns an iterator over `ParserAction`s. A parser action is either a `<script>` tag that needs to be executed or a new encoding that the document should be re-parsed with. The caller can drive the parser by repeatedly advancing this iterator.

The old API is fully preserved, without breaking changes (that I'm aware of).

This is a draft because the design is not final and this needs a companion servo PR to verify the correctness of these changes. Initial feedback is welcome.


simonwuelker:
> The major change is that the https://html.spec.whatwg.org/#input-stream now lives in the html5ever crates (if the decoding wrapper around the tokenizer is used). As a result, the new API surface exposes a "pull" instead of the existing "push" interface.

I am running into architectural issues with this approach, because the prefetch tokenizer needs to run on the same input as the parser, but it shouldn't need to independently decode the incoming bytes. And since the prefetch tokenizer lives in servo, that's going to be difficult when when we decode input in html5ever.
I think it will be easier to keep the decoding in servo.

simonwuelker:
On second thought, perhaps the prefetch tokenizer should live in html5ever too. Currently servo is tokenizing everything two times (on one thread). It would be much cheaper to attach a callback `Fn(&Token)` to the tokenizer that can be used to inspect the tokens that are produced.

simonwuelker:
> The major change is that the https://html.spec.whatwg.org/#input-stream now lives in the html5ever crates (if the decoding wrapper around the tokenizer is used). As a result, the new API surface exposes a "pull" instead of the existing "push" interface.

I think I need to extract the change in a seperate PR first. It's also required for https://github.com/servo/servo/issues/36482 and without it any refactor of the parser structure is more complicated than it needs to be.

simonwuelker:
Closing in favor of https://github.com/servo/html5ever/pull/702

PR #591, opened 2025-04-01
xml5ever: Bubble parser blocking scripts to the caller instead of the TreeSink

simonwuelker:
This is the same approach used by html5ever, which will hopefully allow unifying their API a bit as part of the effort for encoding support.

This is a breaking change. The TreeSink::complete_script method (among other things) is removed.

### DIFF
--- a/xml5ever/src/driver.rs
+++ b/xml5ever/src/driver.rs
@@ -63,7 +63,8 @@ impl<Sink: TreeSink> TendrilSink<tendril::fmt::UTF8> for XmlParser<Sink> {
 
     fn process(&mut self, t: StrTendril) {
         self.input_buffer.push_back(t);
-        self.tokenizer.feed(&self.input_buffer);
+        // FIXME: Properly support </script> somehow.
+        let _ = self.tokenizer.feed(&self.input_buffer);
     }
 
     // FIXME: Is it too noisy to report every character decoding error?
--- a/xml5ever/src/tokenizer/mod.rs
+++ b/xml5ever/src/tokenizer/mod.rs
@@ -23,7 +23,9 @@ use crate::tendril::StrTendril;
 use crate::{buffer_queue, Attribute, QualName, SmallCharSet};
 use log::debug;
 use mac::{format_if, unwrap_or_return};
-use markup5ever::{local_name, namespace_prefix, namespace_url, ns, small_char_set};
+use markup5ever::{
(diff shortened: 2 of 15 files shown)


simonwuelker:
The difference between the way `html5ever` and `xml5ever` block on script elements is best explained by looking at by the `feed` method for both parsers in servo:

The html parser pushes the input into the tokenizer and waits to see if any blocking scripts come back: https://github.com/servo/servo/blob/0d38d6239a4356928c2a7ec0a3cd53ecc32096b0/components/script/dom/servoparser/html.rs#L87-L94

The xml parser pushes the input into the tokenizer, waits until it stops and then checks whether the sink has received any blocking scripts: https://github.com/servo/servo/blob/0d38d6239a4356928c2a7ec0a3cd53ecc32096b0/components/script/dom/servoparser/xml.rs#L45-L51

xml5ever/src/driver.rs
```diff
@@ -63,7 +63,8 @@ impl<Sink: TreeSink> TendrilSink<tendril::fmt::UTF8> for XmlParser<Sink> {
 
     fn process(&mut self, t: StrTendril) {
         self.input_buffer.push_back(t);
-        self.tokenizer.feed(&self.input_buffer);
+        // FIXME: Properly support </script> somehow.
+        let _ = self.tokenizer.feed(&self.input_buffer);
```
simonwuelker, line 67: This is okay, because it is exactly what `html5ever` does too: https://github.com/servo/html5ever/blob/31a2c319c4b9fb763c88fbb7b826e66c3ff372a0/html5ever/src/driver.rs#L107-L109

xml5ever/src/tokenizer/mod.rs
```diff
@@ -631,23 +633,34 @@ macro_rules! go (
...
+    Continue,
+    /// The tokenizer encountered a script element that must be executed
+    /// before tokenization can continue
+    Script(Handle),
+}
+
 impl<Sink: TokenSink> XmlTokenizer<Sink> {
     // Run the state machine for a while.
     // Return true if we should be immediately re-invoked
     // (this just simplifies control flow vs. break / continue).
```
mrobinson, line 661: I guess the doc comment should be updated here.

mrobinson (approved):
Seems good to me, though perhaps someone more familiar with html5ever should review.

jdm (approved):
This is great! Thanks for digging into this.

PR #592, opened 2025-04-04
Bump xml5ever, markup5ever and html5ever versions

simonwuelker:
This should have been done as part of https://github.com/servo/html5ever/pull/591. I forgot that servo depends on the crates.io version of html5ever, not the git repository. So to upgrade in servo we need to release new versions of all three crates.

All three crates have received breaking changes since the last PR.

cc @mrobinson since I think you have permission to create new releases?

### DIFF
--- a/html5ever/Cargo.toml
+++ b/html5ever/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "html5ever"
-version = "0.30.0"
+version = "0.31.0"
 authors = [ "The html5ever Project Developers" ]
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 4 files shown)


jdm (changes requested):
You will need to update https://github.com/servo/html5ever/blob/main/rcdom/Cargo.toml#L19, as well as the markup5ever dependencies in xml5ever and html5ever.

simonwuelker:
Oops, sorry...
Fixed.

jdm (approved):

jdm:
Published.

PR #593, opened 2025-04-06
Don't import enum variants into the global namespace in html5ever

simonwuelker:
Currently the tokenizer and treebuilder in html5ever import all variants of  commonly used enums:

```rust
enum Token {
   FooToken,
   BarToken
}

use Token::*;

fn main() {
    let x = FooToken;
}
```

This is not really idiomatic rust. I've never seen any other codebase do this, and it is not the style we use in servo either.

 This change removes the imports and instead uses the `Token::Foo` notation, renaming the variants where appropriate. 

This is a breaking change, because the `PushFlag` import was public (*even though it really shouldn't have been, I suspect that's another common problem in html5ever*).

### DIFF
--- a/html5ever/src/tree_builder/mod.rs
+++ b/html5ever/src/tree_builder/mod.rs
@@ -35,8 +35,6 @@ use log::{debug, log_enabled, warn, Level};
 use mac::format_if;
 use markup5ever::{expanded_name, local_name, namespace_prefix, namespace_url, ns};
 
-pub use self::PushFlag::*;
-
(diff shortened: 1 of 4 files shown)


mrobinson (approved):
I totally agree! This makes the code a lot easier to follow.

nicoburns (approved):
This does make the code a bit more verbose, but overall this seems like a really good change. Definitely makes the code more readable to me knowing which enum the type is coming from.

Issue #588, opened 2025-03-27
Invalid HTML produces duplicated elements


mohe2015:
The following invalid HTML:
```
<html>
        <body>
        <div class="tb nb"><a href="https://www.humanw.tu-darmstadt.de/fachbereich/index.de.jspZur Homepage des Fachbereichs.</a><br></div>
	<div class="tb" >
    </div>
    </body>
    </html>
```
produces duplicate elements when parsing with `rcdom`:
```
     Running `/home/moritz/Documents/html5ever/target/debug/examples/print-rcdom`
#Document
    <html>
        <head>
        <body>
            #text: \n        
            <div class="tb nb">
                <a href="https://www.humanw.tu-darmstadt.de/fachbereich/index.de.jspZur Homepage des Fachbereichs.</a><br></div>
        <div class=" tb"="">
                    #text: \n    
            <a href="https://www.humanw.tu-darmstadt.de/fachbereich/index.de.jspZur Homepage des Fachbereichs.</a><br></div>
        <div class=" tb"="">
                #text: \n    \n    \n

Parse errors:
    Unexpected token
    Bad character
    Bad character
    Unexpected open element
    Unexpected open tag at end of body
```

simonwuelker:
The second `a` tag is emitted as part of [`reconstructing the active formatting elements`](https://html.spec.whatwg.org/#reconstruct-the-active-formatting-elements). 

> This has the effect of reopening all the formatting elements that were opened in the current body, cell, or caption (whichever is youngest) that haven't been explicitly closed.

I'm not quite sure where it's going wrong yet.

simonwuelker:
The DOM contains no duplicate elements in servo, so I suspect that this is a problem in `rcdom`.

PR #596, opened 2025-04-11
Initialize logger for rcdom examples, cleanup html5ever a bit

simonwuelker:
* Fix some clippy errors
* Add documentation to a few methods
* initialize logger for rcdom examples, so they are more useful for debugging

These are changes that I did while looking a bit into https://github.com/servo/html5ever/issues/588, they're not particularly not important but it feels silly to not commit them.


mrobinson (approved):
Nice!

PR #597, opened 2025-04-12
Don't import xml5ever enums into the global namespace

simonwuelker:
This is mostly the same as https://github.com/servo/html5ever/pull/593 but for xml5ever.


simonwuelker:
CI needs https://github.com/servo/html5ever/pull/596 first

mrobinson (approved):
Thanks for all the great work moving html5ever toward modern Rust!

PR #598, opened 2025-04-12
Remove all direct uses of unwrap_or_return! in html5ever

simonwuelker:
In modern rust this macro can almost always easily be replaced by let-else expressions.

Some uses inside other macros remain, those are not as easy to remove.

### DIFF
--- a/html5ever/src/tree_builder/rules.rs
+++ b/html5ever/src/tree_builder/rules.rs
@@ -329,7 +329,9 @@ where
                     self.unexpected(&tag);
                     if !self.frameset_ok.get() { return ProcessResult::Done; }
 
-                    let body = unwrap_or_return!(self.body_elem(), ProcessResult::Done).clone();
+                    let Some(body) = self.body_elem().map(|b| b.clone()) else {
+                        return ProcessResult::Done;
+                    };
                     self.sink.remove_from_parent(&body);
 
                     // FIXME: can we get here in the fragment case?
--- a/html5ever/src/tokenizer/char_ref/mod.rs
+++ b/html5ever/src/tokenizer/char_ref/mod.rs
@@ -137,19 +137,19 @@ impl CharRefTokenizer {
         tokenizer: &Tokenizer<Sink>,
         input: &BufferQueue,
     ) -> Status {
-        match unwrap_or_return!(tokenizer.peek(input), Stuck) {
-            'a'..='z' | 'A'..='Z' | '0'..='9' => {
+        match tokenizer.peek(input) {
+            Some('a'..='z' | 'A'..='Z' | '0'..='9') => {
                 self.state = Named;
                 self.name_buf_opt = Some(StrTendril::new());
                 Progress
             },
-
-            '#' => {
+            Some('#') => {
                 tokenizer.discard_char(input);
                 self.state = Octothorpe;
                 Progress
             },
-            _ => self.finish_none(),
+            Some(_) => self.finish_none(),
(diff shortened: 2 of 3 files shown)


simonwuelker:
CI needs https://github.com/servo/html5ever/pull/596 first

html5ever/src/tree_builder/rules.rs
```diff
@@ -329,7 +329,9 @@ where
                     self.unexpected(&tag);
                     if !self.frameset_ok.get() { return ProcessResult::Done; }
 
-                    let body = unwrap_or_return!(self.body_elem(), ProcessResult::Done).clone();
+                    let Some(body) = self.body_elem().map(|b| b.clone()) else {
```
mrobinson, line 332: Can this just call `Option::clone()`?

```suggestion
                    let Some(body) = self.body_elem().clone() else {
```

mrobinson (approved):
Excellent!

html5ever/src/tree_builder/rules.rs
simonwuelker: Surprisingly not! 

```
error[E0599]: the method `clone` exists for enum `Option<Ref<'_, Handle>>`, but its trait bounds were not satisfied
    --> html5ever/src/tree_builder/rules.rs:332:55
     |
332  |                     let Some(body) = self.body_elem().clone() else {
     |                                                       ^^^^^ method cannot be called on `Option<Ref<'_, Handle>>` due to unsatisfied trait bounds
     |
    ::: /home/alaska/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:1463:1
     |
1463 | pub struct Ref<'b, T: ?Sized + 'b> {
     | ---------------------------------- doesn't satisfy `Ref<'_, Handle>: Clone`
     |
    ::: /home/alaska/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:571:1
     |
571  | pub enum Option<T> {
     | ------------------ doesn't satisfy `Option<Ref<'_, Handle>>: Clone`
     |
     = note: the following trait bounds were not satisfied:
             `Ref<'_, Handle>: Clone`
             which is required by `Option<Ref<'_, Handle>>: Clone`
note: the method `clone` exists on the type `Ref<'_, Handle>`
    --> /home/alaska/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/clone.rs:165:5
     |
165  |     fn clone(&self) -> Self;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^
help: consider using `Option::expect` to unwrap the `Ref<'_, Handle>` value, panicking if the value is an `Option::None`
     |
332  |                     let Some(body) = self.body_elem().expect("REASON").clone() else {
     |                                                      +++++++++++++++++

For more information about this error, try `rustc --explain E0599`.
error: could not compile `html5ever` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `html5ever` (lib test) due to 1 previous error
```

To be honest, I have no idea what that error is complaining about.

mrobinson: Ah yeah, sometime when there is a `Ref` inside it needs to be dereferenced in order to access the real clone. `Option::as_deref` or `Option::as_ref` might work here, but otherwise the closure is fine.

PR #599, opened 2025-04-13
Split atoms out into a separate crate

nicoburns:
## Motivation

The primary motivation here is shield Stylo (and potentially other crates?) from needing to upgrade to new `markup5ever` versions so frequently. These upgrades are minorly annoying as landing them needs to be synchronised across Servo and Stylo to ensure the versions agree as Stylo uses these types as part of it's public API.

Currently, Stylo has to upgrade with any change to `markup5ever`, however, Stylo only actually uses the Atom types from `markup5ever`, so with this PR breaking changes to to other parts of `markup5ever` (e.g. the TreeSink interface) wouldn't affect Stylo. I believe that adding strings to an Atom type is non-breaking, so this new crate should only rarely require breaking changes.

## Changes made

- Adds a new `web_atoms` crate which splits the `string_cache::Atom` types (and other "data") into it's own crate.
- The types from the new crate are re-exported from `markup5ever` for backwards-compatibility
- The `ns!` macro was change to use a `$crate::` reference to the `namespace_url!` macro. This means that users of the `ns!` don't also need to have the `namespace_url!` in scope.

## Feedback wanted

- Should I also move the `QualName` and `ExpandedName` into the new crate. They are also unlikely to have frequent breaking changes, but they aren't currently required by Stylo

## Potential future extensions

- We could consider merging the `stylo_atoms` crate into this new crate. Then we could manage all the string_cache codegen in one place, and it would be less code to maintain in a repository that is synced from an upstream.


jdm:
I haven't looked closely at the changes yet, but I'm definitely in favour of reducing dependencies on markup5ever.

jdm (approved):
Let's do it! I'm not sure I see a good reason to move QualName and ExpandedName right now.

nicoburns:
This Servo and Stylo PRs are both passing CI, although it is causing a lint failure (an import that was previously necessary is no longer necessary - this doesn't break the build but it does trigger a lint) for Servo.

So I think we could release this as an initial release of `web_atoms` + a patch release of `markup5ever`, but we'd probably want to upgrade Servo to the new version straight away (switching Stylo to the new crate could be done asynchronously).

PR #600, opened 2025-04-13
Trigger errors without macros, remove `unwrap_or_else!` and `go_match!`

simonwuelker:
This change is the first step in gradually removing the `go!` macro that is currently used to implement the tokenizer-DSL.

When this macro was first introduced it might have made sense to have it - some quality-of-life rust syntax did not exist back then.

But I think it's not necessary anymore and I'm certain that it's possible to remove the DSL without compromising on performance or readability (like [this](https://github.com/simonwuelker/Stormlicht/blob/863ec3b841424629a8356b8786f60e305d95a877/crates/web/src/html/tokenization/tokenizer.rs#L1341-L1361)).

### DIFF
--- a/html5ever/src/tokenizer/mod.rs
+++ b/html5ever/src/tokenizer/mod.rs
@@ -658,13 +658,6 @@ macro_rules! go (
     ( $me:ident : ) => (());
 );
 
-macro_rules! go_match ( ( $me:ident : $x:expr, $($pats:pat),+ => $($cmds:tt)* ) => (
-    match $x {
-        $($pats)|+ => go!($me: $($cmds)*),
-        _ => (),
-    }
-));
-
 // This is a macro because it can cause early return
 // from the function where it is used.
 macro_rules! get_char ( ($me:expr, $input:expr) => (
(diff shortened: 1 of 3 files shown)


html5ever/src/tokenizer/mod.rs
```diff
@@ -371,7 +371,11 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
         TokenizerResult::Done
     }
 
+    #[inline]
     fn bad_char_error(&self) {
+        #[cfg(feature = "trace_tokenizer")]
+        trace!("  error");
```
simonwuelker, line 377: I think the usefulness of the `trace_tokenizer` feature is a bit questionable, but I didn't want to touch it for now. This matches what the `sh_trace!` macro would log before.

html5ever/src/tokenizer/mod.rs
```diff
@@ -1565,7 +1578,10 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
 
             states::MarkupDeclarationOpen => go!(self: error; to BogusComment),
 
-            states::CdataSection => go!(self: emit_temp; error_eof; to Data),
+            states::CdataSection => {
+                self.bad_eof_error();
```
jdm, line 1745: This reorders the eof and emit_tmp. Is that a problem?

jdm (approved):

html5ever/src/tokenizer/mod.rs
simonwuelker: Any implementation that relies on this ordering is doing silly things, but it is observable. Nice catch!

PR #610, opened 2025-05-06
Add missing event handler names

KiChjang:
All of them are from https://html.spec.whatwg.org/multipage/webappapis.html#idl-definitions.


KiChjang:
Hmm... it looks like I don't actually need to add all of these, the specific ones that are really important to have are the ones in https://html.spec.whatwg.org/multipage/#event-handlers-on-elements,-document-objects,-and-window-objects:event-handlers-6.

mrobinson (approved):

mrobinson:
Published: https://crates.io/crates/web_atoms/0.1.2

PR #620, opened 2025-06-19
Publish html5ever 0.32.0.

jdm:
This version includes https://github.com/servo/html5ever/pull/593, which is a breaking change.


simonwuelker (approved):
You need to update the html5ever version in the other workspace packages.

PR #621, opened 2025-06-20
Publish match_token 0.1.1

jdm:
This version includes a change that's required to build the newest version of html5ever.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,5 @@ resolver = "2"
 rust-version = "1.70.0"
 
 [workspace.dependencies]
-match_token = { version = "0.1", path = "match_token" }
+match_token = { version = "0.1.1", path = "match_token" }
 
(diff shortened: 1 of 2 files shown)


mrobinson (approved):

Quentin01:
Just so you know @jdm, as you published this ~45 minutes ago, you broke the install of `mdbook` as you didn't yet published the new version of `html5ever` and `markup5ever` yet. I hope you will publish those new versions ASAP.

This version contains a breaking change from #593, it should have been a bump in the minor version instead of the patch.

jdm:
@Quentin01 I apologize for the problems. This all started with attempting to release a new version of html5ever and discovering multiple that publishing was broken while normal cargo builds worked fine. We're going to sort out a better system to avoid this in the future.

nicoburns:
@jdm I suspect this problem arising means that the new version of `match_token` is semver breaking and ought to be published as 0.2 (with 0.1.1 yanked).

(even once the new version of `html5ever` is published the old version ought to continue to work)

Quentin01:
I opened an issue before seeing your comments. I should have probably waited more, but we can probably use it to discuss this issue further. I'm glad to know you are on it and that you will sort this out. Thanks.

PR #635, opened 2025-06-26
Update version number in Readme

atouchet:


simonwuelker (approved):
Apologies, I'll remember to bump this in the future... 

Maybe we should just reference `cargo add html5ever`.

nicoburns:
Yeah, we should probably just remove the version number from the README. Cargo add is easier for users anyway :)

mrobinson:
I think this isn't necessary any longer.

PR #640, opened 2025-06-30
Replace `match_token!` proc macro with `tag!` macro_rules macro

nicoburns:
Replaces the `match_token!` procedural macro to be replaced with a series of simpler `tag!` macro_rules macros combined with a regular rust `match` expression.

Requires a new string-cache-codegen release:
- https://github.com/servo/string-cache/pull/297

The following code block:

```rust
//§ the-before-html-insertion-mode
InsertionMode::BeforeHtml => match_token!(token {
    Token::Characters(SplitStatus::NotSplit, text) => ProcessResult::SplitWhitespace(text),
    Token::Characters(SplitStatus::Whitespace, _) => ProcessResult::Done,
    Token::Comment(text) => self.append_comment_to_doc(text),

    tag @ <html> => {
        self.create_root(tag.attrs);
        self.mode.set(InsertionMode::BeforeHead);
        ProcessResult::Done
    }

    </head> </body> </html> </br> => else,

    tag @ </_> => self.unexpected(&tag),

    token => {
        self.create_root(vec!());
        ProcessResult::Reprocess(InsertionMode::BeforeHead, token)
    }
}),
```

would be replaced with:

```rust
// § the-before-html-insertion-mode
// <https://html.spec.whatwg.org/multipage/parsing.html#the-before-html-insertion-mode>
InsertionMode::BeforeHtml => {
    let anything_else = |token: Token| {
        self.create_root(vec![]);
        ProcessResult::Reprocess(InsertionMode::BeforeHead, token)
    };

    match token {
        Token::Comment(text) => self.append_comment_to_doc(text),

        Token::Characters(SplitStatus::NotSplit, text) => {
            ProcessResult::SplitWhitespace(text)
        },

        Token::Characters(SplitStatus::Whitespace, _) => ProcessResult::Done,
        Token::Tag(tag @ tag!(<html>)) => {
            self.create_root(tag.attrs);
            self.mode.set(InsertionMode::BeforeHead);
            ProcessResult::Done
        },

        Token::Tag(tag!(</head> | </body> | </html> | </br>)) => anything_else(token),
        Token::Tag(tag @ tag!(</>)) => self.unexpected(&tag),

        token => anything_else(token),
    }
},
```


simonwuelker (commented):
I like the idea!

Having to write `tag_token!(tag @ ["html"] | [/"body"])` is odd though.
Can't it be `tag_token!(tag @ <html> | </body>)` if the macro match arm is changed from `(__inner:[/$tag:tt])` to something like `(__inner:</$tag:ident>)`?

You could convert the `ident` to a `tt` using the [`stringify`](https://doc.rust-lang.org/std/macro.stringify.html) macro and that would remove the need for quotes at the macro invocation site.

nicoburns:
@simonwuelker I managed to get <> instead of [] working. I also got bare <> and </> working. But I couldn't get rid of the quotes around the tag name.

The problem is order of evaluation. You can't do `local_name!(stringify!($tag))` because `local_name!` gets passed the literal tokens `stringify!($tag)` rather the evaluation of that macro.

I think it could be made to work if we either made `local_name` take bare unquoted token or we create a `const fn` to create a const `LocalName`.

nicoburns:
Ok, I've made this work. We can't use `const fn` until https://github.com/rust-lang/rust/issues/76001 lands, but I got it working by making the string_cache macros accept idents for any atom that is entirely alphanumeric+underscores (https://github.com/servo/string-cache/pull/296).

I've updated the example in PR description with the new syntax.

nicoburns:
@simonwuelker @jdm (no urgency but:)

This is now fully working and is in a position where it could be implemeted. It would just be a case of going through the `rules.rs` file and updating every case to use the new syntax (this PR currently updates the first 2 cases to give an idea of what that would look like). So it would be good to get your views on whether this is a change we want to pursue.

The main difference are:
- This would remove the need for the `match_token` proc macro and crate entirely, replacing it with the `macro_rules` macros in this PR (this also creats some minor formatting differences (see `Token::Characters(SplitStatus::NotSplit` cases) because auto-formatting now applies to these code blocks which it didn't before because of the macro).
- The syntax in this PR is a little more verbose, but also less magic. The macro_rules macro outputs a pattern which then allows the match block to be a regular rust `match` rather than being processed by the proc_macro.
- The "else case" is the other thing that changes here. Rather than the special "else syntax" like `</head> </body> </html> </br> => else` (which causes cases that match it to fall through to the fallback case) you have a guard like `if is_not_tag!(tag, </head> | </body> | </html> | </br>)` (or equivalently `if !matches!(tag, tag!(</head> | </body> | </html> | </br>))`. Again, this is a lot less magic and works like standard rust. But it slightly less directly follows the order of the spec as the guard has to be moved down to the "every start tag" and/or "every end tag case" rather than being specified further up (which the spec does for some reason). We could perhaps leave a comment where the "else" case is in the spec that refers you down to the guard on the later clause?)
- For example of spec, see https://html.spec.whatwg.org/multipage/parsing.html#the-before-head-insertion-mode for example of spec. Either way, I think we could benefit from copying more of the spec into the rules.rs file.

Thoughts?

P.S. Minor note on syntax: it would also be possible to write `TagToken!(tag @ <html>)` instead of `Token::Tag(tag @ tag!(<html>))` instead of  if we preferred.

simonwuelker:
> The "else case" is the other thing that changes here. Rather than the special "else syntax" like </head> </body> </html> </br> => else (which causes cases that match it to fall through to the fallback case) you have a guard like if is_not_tag!(tag, </head> | </body> | </html> | </br>) (or equivalently if !matches!(tag, tag!(</head> | </body> | </html> | </br>)). Again, this is a lot less magic and works like standard rust. But it slightly less directly follows the order of the spec as the guard has to be moved down to the "every start tag" and/or "every end tag case" rather than being specified further up (which the spec does for some reason). We could perhaps leave a comment where the "else" case is in the spec that refers you down to the guard on the later clause?)

Personally I would prefer to implement functions for the "else" cases (like `any_other_tag_in_data_state`) and the simply call those in two places.

Other than that, this looks great! I'm all for less magic and less proc macros, even if it ends up being slightly more verbose.

> P.S. Minor note on syntax: it would also be possible to write TagToken!(tag @ <html>) instead of Token::Tag(tag @ tag!(<html>)) instead of if we preferred.

Not a big difference but I prefer to keep the macro invocation as small as possible(the current version).

> Either way, I think we could benefit from copying more of the spec into the rules.rs file.

Yeah adding proper spec comments is something else that I eventually want to come around too... But it's a dull task (:

jdm:
>Other than that, this looks great! I'm all for less magic and less proc macros, even if it ends up being slightly more verbose.

Same! I love removing syntax unique to this crate.

nicoburns:
Ok, I've rolled this out to the entire `rules.rs`, cleaned it up, documented the `tag!` macro, and deleted the `match_token` crate. I've gone with a closure for the `else` cases (which actually only occur in 5 of the `InsertionMode`s). Normally I would a function for this, but writing out the generic types (which can't be reused from the outer scope) made that verbose in this case).

I haven't added proper spec comments, but I have added links to the spec for each `InsertionMode`. As part of this, I noticed two insertion modes ("in select" and "in select in table") which didn't appear to be the spec. Turns out the spec has changed and these modes no longer exist. I've opened https://github.com/servo/html5ever/issues/660 to track this.

~~We'll need a release of `string-cache-codegen` (https://github.com/servo/string-cache/pull/297) before this can be merged, but otherwise this is ready to go (modulo any further review feedback~~ (now published)

html5ever/src/tree_builder/rules.rs
```diff
@@ -34,19 +33,74 @@ fn current_node<Handle>(open_elems: &[Handle]) -> &Handle {
...
+    /// <https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inhtml>
     pub(crate) fn step(&self, mode: InsertionMode, token: Token) -> ProcessResult<Handle> {
         self.debug_step(mode, &token);
 
         match mode {
-            //§ the-initial-insertion-mode
-            InsertionMode::Initial => match_token!(token {
-                Token::Characters(SplitStatus::NotSplit, text) => ProcessResult::SplitWhitespace(text),
+            // § The "initial" insertion mode
+            // <https://html.spec.whatwg.org/multipage/parsing.html#the-initial-insertion-mode>
```
simonwuelker, line 99: It's a bit redundant to have both the spec link and the paragraph title. Also this is a regular comment, so the `<>` around the link are not required.

This is a small nit, feel free to keep it this way if it's too much effort to change all these references again :D

simonwuelker (approved):

html5ever/src/tree_builder/rules.rs
nicoburns: I've removed the `<>` (easy with multiple cursors), but I've left the text labels as some of them differ from the link fragment (e.g. `§ The "text" insertion mode` and `https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-incdata`)

PR #643, opened 2025-07-14
Simplify the way character references are resolved

simonwuelker:
Currently the `CharRefTokenizer` has a `result` field. When a reference is fully resolved then the `result` field is set, and `Status::Done` is returned. The caller can then call `CharRefTokenizer::get_result` to obtain the resulting character reference. It is easier to instead return the reference directly as a field in `Status::Done`.

Additionally, this PR resolves a TODO about handling of the char ref tokenizer in the html tokenizer, allowing the field to no longer be boxed.

Not a breaking change - none of this publicly exposed.

### DIFF
--- a/html5ever/src/tokenizer/char_ref/mod.rs
+++ b/html5ever/src/tokenizer/char_ref/mod.rs
@@ -55,6 +55,13 @@ pub(super) struct CharRefTokenizer {
     name_len: usize,
 }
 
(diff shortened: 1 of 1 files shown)


simonwuelker:
Passing try run: https://github.com/simonwuelker/servo/actions/runs/16271092452

PR #645, opened 2025-07-23
Tokenize named character references using a DAFSA

simonwuelker:
Currently named character references are implemented using a `phf` map that is repeatedly queried for each character. This works, but has suboptimal performance and a significant impact on binary size.

Traversing a DAFSA that is generated at compile time makes tokenizing named character references 30% faster. This technique is described in https://www.ryanliptak.com/blog/better-named-character-reference-tokenization/. For illustration, a reduced version of the dafsa is can be viewed [here](https://mermaid.live/edit#pako:eNqNlsuO2kAQRX_F6kVWGLnfZUeK8t4lyiKriI0HezAKtpExUQji38MguV2I1tXsMBxX163Hbc5i3Ve1KESSbIZy3yQ_P666JFmN369fZ-_PyaEp93WRrLfD-pKmyZskTd-9_CZnTD5iHyZMzZh6xL5MmIbYtwkjiJUhNwm5p8B5yK2DCJzec1CRQW4TOBxvF7gccu3EGQO5LnA4Xj9xFndtP3EO6x0mzmO9h8Dh_o5hDFjfyD-C2wCygESg0sQqQ_kjWIWRYaWW5hGsQ21mzs1Y9bRLJ_RtkD2jnsAY8vGKgaGOxPpHCtWHNYY0mFjLZFsDRkzx1YsYQ1S0Bil6drJHJ1uWIVDC-5Kh2WZCnASj41ixXaTYpwCyYruI5K8BZJJdRPIxgEyzs6A2jqt2YMQca4vzYAUd2yxHoH-WqbYamQ4fMeSedxGRPVkmxnrg23xoDdr-O9CAG8iwzhj7yp3WORBjGGgig9sEkOcoQY5cjNbAvjXrjEaqNVOtI6r_xFxUZ-DoOztBNqEMBk_RiAYdzcQo1EK2W9KDgJKtjCRwdUh2GckclFHx-zcDEbloKYEWycxMKmA9ki2h1MAoeKsjon9EqqiB27JoBrmOWIjNsK1EMQ7HeiHaemjLl0dxvgUQY1O39UoU149VOfxeiVV3ub6zL7tffd9Orw39cdOI4rncHa5Px31VjvXnbXn9pzwjdVfVw6f-2I2iUORvMURxFn9FobOlzYk0UeakVdqYhTiJIpVeLcm6zJAlnZPRdFmIf7dz5VKR9Npo5XPltbLm8h8zQpsS).

Apologies for the big change. If it's too hard to review we could also merge the DAFSA incrementally. Most of the diff is the list of named entities being moved around and a benchmark file being added.

I have not looked into how this affects the binary size. Some more savings are possible by packing the array of result characters.

Do not merge yet - needs servo companion PR.

This is a breaking change for `markup5ever` and `web_atoms`.

### DIFF
--- a/RELEASING.MD
+++ b/RELEASING.MD
@@ -21,6 +21,13 @@ The **web_atoms** crate is on a separate cycle as it needs frequent releases but
 - Publish the new version of **web_atoms**
 - Optionally: publish a new version of the other crates to match
 
+## Making a release of **web_atoms**:
+
+- Bump the version in `html_named_entities/Cargo.toml`
+- Update the version **html_named_entities** in the workspace `Cargo.toml`'s `[workspace.dependencies]` section to match
+- Publish the new version of **html_named_entities**
+- Optionally: publish a new version of the other crates to match
+
 ## Making a release of all other crates
 
 In the workspace `Cargo.toml`:


nicoburns:
Would it be possible to separate out the DAFSA so that it doesn't depend on markup5ever traits? A small/fast HTML entity parser (possibly following a "sans-io" design?) seems like a really useful standalone crate.

btw, is a "DAFSA" the same thing as what burntsushi calls a "DFA" in his blogposts on the `regex` crate?

simonwuelker:
> Would it be possible to separate out the DAFSA so that it doesn't depend on markup5ever traits? A small/fast HTML entity parser (possibly following a "sans-io" design?) seems like a really useful standalone crate.

Yeah sure, I wasn't (and am still not) sure if adding another crate is worth it. Though it would likely only require very little maintenance, because the list of named entities will never change...

simonwuelker:
> btw, is a "DAFSA" the same thing as what burntsushi calls a "DFA" in his blogposts on the regex crate?

Pretty much, yes -  a DFA is a *deterministic finite state automaton*. DAFSAs are a subset of that, which are acyclic. (DAFSA  = *deterministic acyclic finite state automaton*).

html_named_entities/build/main.rs
```diff
@@ -0,0 +1,373 @@
...
+    codepoints: Vec<u32>,
+    #[allow(dead_code)]
+    characters: String,
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=build/");
+
+    // Parse the list of named entities from https://html.spec.whatwg.org/entities.json
+    let input_file = BufReader::new(File::open("build/entities.json").unwrap());
```
nicoburns, line 259: Consider converting the JSON file into a Rust file. Then you wouldn't need to depend on `serde` in the build.rs.

html_named_entities/src/codegen.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use super::CharRef;
+
+include!(concat!(env!("OUT_DIR"), "/named_entities_graph.rs"));
+
+/// A single node in the DAFSA.
+///
+/// For memory efficiency reasons, this is packed in 32 bits. The memory representation is as follows:
+/// * 8 bits: code point
+/// * 8 bits: hash value
```
nicoburns, line 18: This is missing part of the memory layout

PR #647, opened 2025-07-24
Generate rustdoc documentation for all publicly exported items

simonwuelker:
A few of the items in xml5ever are annotated with `#[doc(hidden)]`. This causes the `initial_state` field in https://docs.rs/xml5ever/latest/xml5ever/tokenizer/struct.XmlTokenizerOpts.html to not link to the enum, which is bad.

Presumably this is done because the author wanted to get around the `#[deny(missing_docs)]` lint? I'm not sure. Not all of these need to be public and I have reduced the visibility of a few items (`NamespaceMap` and a few methods) to `pub(crate)`. This makes this PR a breaking change. I believe that these were not intended to be public in the first place, but I can undo that if requested.

### DIFF
--- a/xml5ever/src/tokenizer/states.rs
+++ b/xml5ever/src/tokenizer/states.rs
@@ -12,19 +12,19 @@
 //! This is public for use by the tokenizer tests.  Other library
 //! users should not have to care about this.
 
-pub use self::AttrValueKind::*;
-pub use self::DoctypeKind::*;
-pub use self::XmlState::*;
+#![allow(missing_docs)] // FIXME
+
+pub use AttrValueKind::*;
+pub use DoctypeKind::*;
+pub use XmlState::*;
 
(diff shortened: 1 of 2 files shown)


nicoburns (commented):
I believe `doc(hidden)` was used here because these types were/are not intended to be part of the public API, but needed to be `pub` so that tests could access them.

simonwuelker:
Possible, either way, they're not needed anymore (tests run fine without them)

xml5ever/src/tokenizer/states.rs
```diff
@@ -12,19 +12,19 @@
 //! This is public for use by the tokenizer tests.  Other library
 //! users should not have to care about this.
 
-pub use self::AttrValueKind::*;
-pub use self::DoctypeKind::*;
-pub use self::XmlState::*;
+#![allow(missing_docs)] // FIXME
```
mrobinson, line 15: Please open a bug for this and link to it here. Thanks!

mrobinson (approved):

PR #653, opened 2025-07-27
Remove lint allow statements for markup5ever

simonwuelker:
Neither of these are relevant anymore. `#![allow(unexpected_cfgs)]` was only necessary because of some legacy code for a feature that doesn't exist (anymore?).


nicoburns (approved):
Yes, `heap_size` was replaced with `malloc_size_of` (https://github.com/servo/malloc_size_of) many years ago.

PR #655, opened 2025-08-15
ci: Integrate github inline annotation for clippy

jerensl:
Recently, we decided to centralize reusable actions in github/actions to make maintenance and updates easier across repositories. As part of this change, we are now adding Clippy annotations to some repos to make it easier to view errors directly in PRs.

https://github.com/servo/ipc-channel/pull/405#discussion_r2247738361

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,7 +71,9 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Run clippy
-        run: cargo clippy --all-features --all-targets -- -D warnings
+        uses: servo/actions/cargo-annotation@main
+        with:
+          cargo-command: clippy --all-features --all-targets -- -D warnings
 
   build_result:
     name: Result


.github/workflows/main.yml
```diff
@@ -71,7 +71,9 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Run clippy
-        run: cargo clippy --all-features --all-targets -- -D warnings
+        uses: servo/actions/cargo-annotation@main
+        with:
+          cargo-command: clippy --all-features --all-targets
```
mukilan, line 76: Is it correct that action will only fail the build as long as annotations are enabled? If annotations are ever disabled using `with-annotations: false` then only `cargo clippy` will be run *but* without the `-D warnings` that this CI setup is using. Is there a way we can fail the build even if annotations are disabled?

jerensl: We can still use -D warnings when with-annotations is set to false. This flag only causes issues if --message-format=json is specified after -- -D warnings. It only works when placed before it.

mukilan: Not sure I follow. The issue I'm referring to is the behaviour of `cargo clippy` when `-D warnings` is not passed (as in our action). Without `-D warnings`, `cargo clippy` will exit successfully even when there are lint violations and won't block the build.

In the current setup in this repo, `-D warnings` is being passed explicitly to make `cargo clippy` fail when there are violations. But in our action, we don't do that and rely on the annotations.sh to `exit 1` if there are violations. However, that doesn't happen if annotations get disabled.

The question is if we can change the action to always fail, even if `with-annotations` is false.

mukilan:
> Testing: https://github.com/jerensl/html5ever/actions/runs/16986748929/job/48157192732?pr=1#step:6:199

@jerensl have you tested this change with the recent changes to servo/actions/cargo-annotations as well?

jerensl:
> > Testing: https://github.com/jerensl/html5ever/actions/runs/16986748929/job/48157192732?pr=1#step:6:199
> 
> @jerensl have you tested this change with the recent changes to servo/actions/cargo-annotations as well?

I haven't tested on this repo as the PR just merged yet, but im testing that on other [repo here](https://github.com/jerensl/ipc-channel/pull/2) before it getting merged

mukilan:
I guess we can trigger another test run from a fork of this repo now that more changes have been merged to servo/actions.

jerensl:
For the new changes in `servo/action`, I been testing on [this pr](https://github.com/jerensl/html5ever/pull/2), and [here](https://github.com/jerensl/html5ever/actions/runs/17286983289/job/49065906417?pr=2#step:6:1) is the result from CI

mukilan:
> For the new changes in `servo/action`, I been testing on [this pr](https://github.com/jerensl/html5ever/pull/2), and [here](https://github.com/jerensl/html5ever/actions/runs/17286983289/job/49065906417?pr=2#step:6:1) is the result from CI

Do we have a run to validate the failure scenario as well?

jerensl:
> > For the new changes in `servo/action`, I been testing on [this pr](https://github.com/jerensl/html5ever/pull/2), and [here](https://github.com/jerensl/html5ever/actions/runs/17286983289/job/49065906417?pr=2#step:6:1) is the result from CI
> 
> Do we have a run to validate the failure scenario as well?

[Here](https://github.com/jerensl/html5ever/pull/1) the PR for failure scenario when there a warning in clippy and the [result](https://github.com/jerensl/html5ever/actions/runs/17319858395/job/49170506865?pr=1#step:6:58)

mukilan:
Could you trigger a push to the test PR again after https://github.com/servo/actions/pull/7 lands?

mukilan:
Also the Github Annotation doesn't have the description of the error i.e:

```
   --> html5ever/src/serialize/mod.rs:82:23
   |
82 |         let numbers = vec![1, 2, 3, 4];
   |                       ^^^^^^^^^^^^^^^^ help: you can use an array directly: `[1, 2, 3, 4]`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_vec
   = note: `#[warn(clippy::useless_vec)]` on by default
   ```
   
   It just displays the the message title again in the body.

**Comment by jerensl** [comment]
> Also the Github Annotation doesn't have the description of the error i.e:
> 
> ```
>    --> html5ever/src/serialize/mod.rs:82:23
>    |
> 82 |         let numbers = vec![1, 2, 3, 4];
>    |                       ^^^^^^^^^^^^^^^^ help: you can use an array directly: `[1, 2, 3, 4]`
>    |
>    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_vec
>    = note: `#[warn(clippy::useless_vec)]` on by default
> ```
> 
> It just displays the the message title again in the body.

My bad, I forget the escape newlines https://github.com/servo/actions/pull/8

**Comment by mukilan** [comment]
Looks like the violations are counted twice - once as 'warning' and once as 'error' - both in the workflow's log and the GH annotations.

**Comment by jerensl** [comment]
> Looks like the violations are counted twice - once as 'warning' and once as 'error' - both in the workflow's log and the GH annotations.

That's come from clippy because we have deny warning in this file [here](https://github.com/servo/html5ever/blob/main/html5ever/src/lib.rs#L12C23-L12C24)

Some Clippy lints are categorized as notes rather than warnings, do we also need to handle those with inline annotations?
```
note: the lint level is defined here
  --> html5ever/src/lib.rs:12:24
   |
12 | #![cfg_attr(test, deny(warnings))]
   |                        ^^^^^^^^
   = note: `#[deny(clippy::len_zero)]` implied by `#[deny(warnings)]`

```

mukilan:
> > Looks like the violations are counted twice - once as 'warning' and once as 'error' - both in the workflow's log and the GH annotations.
> 
> That's come from clippy because we have deny warning in this file [here](https://github.com/servo/html5ever/blob/main/html5ever/src/lib.rs#L12C23-L12C24)

The real cause seems to be the `--all-targets` flag we pass to clippy, which runs clippy for both the library and the test targets and since the `deny(warning)` is gated on `cfg(test)`, clippy uses one level for library target and another for test target, resulting in two issues.

The `deny(warning)` was [gated on tests to allow dependants to build](https://github.com/servo/html5ever/commit/ec7e7e94b1b46de9bf8d2512eb6b17f277897bc7) but I think this is not necessary any more as I think cargo will anyway cap the lints from dependencies. However, that is a separate change and not needed for this to land.

> 
> Some Clippy lints are categorized as notes rather than warnings, do we also need to handle those with inline annotations?
> 
> ```
> note: the lint level is defined here
>   --> html5ever/src/lib.rs:12:24
>    |
> 12 | #![cfg_attr(test, deny(warnings))]
>    |                        ^^^^^^^^
>    = note: `#[deny(clippy::len_zero)]` implied by `#[deny(warnings)]`
> ```


I don't think we need to handle this.

mukilan (approved):

PR #656, opened 2025-08-19
match_token: add missing license texts

decathorpe:
I came across this issue when preparing package updates for html5ever et al. for Fedora.

It looks like when match_token was split off into a separate crate (in https://github.com/servo/html5ever/commit/feced062a9119d6f8279cf9d002f9a93926a8ca4), adding the Apache-2.0 and MIT license texts was missed.

On the other hand, this was done correctly when when the web_atoms crate was split off (in https://github.com/servo/html5ever/commit/0292f3fddfe2926aa06c9b67c5ed312a67047ec3) so the fact that the license texts are missing from the match_token crate looks like a simple accidental omission.

### DIFF
--- a/match_token/LICENSE-APACHE
+++ b/match_token/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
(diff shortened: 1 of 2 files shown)


nicoburns (approved):

nicoburns:
`match_token` is definitely covered by the repo-level Apache-2.0 and MIT licenses, but I see no reason not to add license files to the crate as well.

decathorpe:
Thank you!

PR #667, opened 2025-09-09
Impl `PartialEq<str>` for `Tendril<ASCII>` and `Tendril<UTF8>`

nicoburns:
- Fixes https://github.com/servo/tendril/issues/30

### DIFF
--- a/tendril/src/tendril.rs
+++ b/tendril/src/tendril.rs
@@ -22,7 +22,7 @@ use encoding::{self, DecoderTrap, EncoderTrap, EncodingRef};
 
 use crate::buf32::{self, Buf32};
 use crate::fmt::imp::Fixup;
-use crate::fmt::{self, Slice};
+use crate::fmt::{self, Slice, ASCII, UTF8};
 use crate::util::{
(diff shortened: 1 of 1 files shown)


simonwuelker:
Feels a bit weird to compare tendril bytes without knowing what format they are. This will yield unexpected results when comparing `str` with tendrils using non-utf8 encodings.

Maybe we could only impl this for the formats that come with `tendril` out of the box(https://docs.rs/tendril/latest/tendril/fmt/index.html#structs) ? Those that are known to be subsets of utf8.

nicoburns:
> Feels a bit weird to compare tendril bytes without knowing what format they are. This will yield unexpected results when comparing `str` with tendrils using non-utf8 encodings.
> 
> Maybe we could only impl this for the formats that come with `tendril` out of the box(https://docs.rs/tendril/latest/tendril/fmt/index.html#structs) ? Those that are known to be subsets of utf8.

I'm inclined to agree with this. I've restricted the `impl` to `Tendril<ASCII>` and `Tendril<UTF8>`.

I think one could argue for an impl for `Tendril<WTF8>` too. And a comparision to `&[u8]` for Bytes. But I suspect the UTF8 one is the most useful, and we can always add more later.

PR #683, opened 2025-10-28
Continue parsing instead of stopping on script in xml5ever.

halli2:
Changes `XmlParser::process` to function like the html equivelant.

I think this is what it was meant to be when it was implemented here: https://github.com/servo/html5ever/pull/591#pullrequestreview-2734323784


simonwuelker:
Sorry, I forgot about this PR: If you rebase then CI should be green.

PR #684, opened 2025-10-28
Bump MSRV to 1.71

simonwuelker:
Our MSRV is 1.70, parking_lot 0.12.5 needs 1.71. I was not able to tell cargo to pick compatible versions for both `parking_lot`, `parking_lot_core` and `lock_api`. (We can pin `parking_lot` in `Cargo.toml` but that seems to make cargo ignore the specified versions for the other two)

1.71 was released in summer of 2023 so upgrading should not be a problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Install stable toolchain
         run: |
           rustup set profile minimal
-          rustup override set 1.70.0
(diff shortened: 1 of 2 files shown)


simonwuelker:
Unblocks CI for https://github.com/servo/html5ever/pull/683

simonwuelker:
https://github.com/Amanieu/parking_lot/issues/470#issuecomment-2923514115

PR #689, opened 2025-11-05
Add oncommand and onbeforetoggle local names

lukewarlow:
Missed these from the previous PRs


lukewarlow:
Out of interest how do I actually use these within Servo? Do I need to point to a new commit or wait for a release?

nicoburns:
> Out of interest how do I actually use these within Servo? Do I need to point to a new commit or wait for a release?

Generally you'd start a Servo PR by pointing to a git commit. Then for this crate we might try to do a release before actually merging that PR.

PR #654, opened 2025-08-06
Update phf to 0.13

xu-cheng:
The phf in string-cache is updated at https://github.com/servo/string-cache/pull/295, which this PR depends on.

PR #690, opened 2025-11-11
Bump versions

Loirooriol:
Bumps html5ever, markup5ever and xml5everf to 0.36.0.
Bumps web_atoms to 0.1.4.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.35.0"
+version = "0.36.0"
 license = "MIT OR Apache-2.0"
 authors = [ "The html5ever Project Developers" ]
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):

mrobinson:
These have been published by @emilio:
 - https://crates.io/crates/html5ever/0.36.0
 - https://crates.io/crates/web_atoms/0.1.4

xu-cheng:
Hi, according to [this](https://github.com/servo/html5ever/blob/main/RELEASING.MD#making-a-release-of-web_atoms), it seems that `web_atoms` should be bumped to 0.2.0 instead due to `phf` breaking change from #654. The 0.1.4 release may need to be yanked.

Loirooriol:
Even if phf had potentially breaking changes, I don't think that web_atoms had any, so I thought that a 0.1.4 was enough?

xu-cheng:
I’m actually not entirely sure what is the correct answer.

The question is what would happen if old html5ever release pulls the latest web_atom of which a breaking release phf is used underneath. The breaking release phf may cause trait conflicts or inconsistent hash being used. But since html5ever itself does not depend on phf directly. It may be fine? 🤷‍♂️

Loirooriol:
No problem from my side if you want to bump to 0.2.0

jdm:
Yes, we'll need to yank and publish 0.2.0 instead.

nicoburns:
I can confirm that the Blitz build does actually break if I run `cargo update` (which causes it to pick up `web_atoms` 0.1.4), so we definitely need to republish and yank this release. I've put up https://github.com/servo/html5ever/pull/691 with new versions.

Issue #696, opened 2025-11-30
serialize() should escape '<' and '>' in attributes, according to the HTML Standard


pizzacat83:
The current implementation of `html5ever::serialize::serialize` does not escape the characters `<` and `>` within attribute values.

Historically this was the behavior required by the HTML Standard, but recently there was [a change in the HTML Standard](https://github.com/whatwg/html/commit/e21bd3b4a94bfdbc23d863128e0b207be9821a0f) to mandate escaping `<` and `>` in attributes, to mitigate a class of XSS attacks.

As a result, the output of `html5ever::serialize::serialize` does not conform to the latest HTML Standard.

### Expected behavior

`<span id="<>"></span>` should be serialized as `<span id="&lt;&gt;"></span>`.

The following test (for `rcdom/tests/html-serializer.rs`) demonstrates this.

```rust
test!(
    attribute_ltgt,
    r#"<span id="<>"></span>"#,
    r#"<span id="&lt;&gt;"></span>"#
);
```

However, the above test fails in the current implementation, and the actual serialized output is `<span id="<>"></span>`.

Note that Firefox ([release note](https://www.firefox.com/en-US/firefox/140.0/releasenotes/)) and Chrome ([release note](https://developer.chrome.com/blog/escape-attributes)) adhere to this expected behavior.

## Proposed fix

By applying the patch below, the implementation becomes conformant with the above change of the HTML Standard, and the above test passes.

```diff
diff --git a/html5ever/src/serialize/mod.rs b/html5ever/src/serialize/mod.rs
index 710066a..eddca78 100644
--- a/html5ever/src/serialize/mod.rs
+++ b/html5ever/src/serialize/mod.rs
@@ -107,8 +107,8 @@ impl<Wr: Write> HtmlSerializer<Wr> {
                 '&' => self.writer.write_all(b"&amp;"),
                 '\u{00A0}' => self.writer.write_all(b"&nbsp;"),
                 '"' if attr_mode => self.writer.write_all(b"&quot;"),
-                '<' if !attr_mode => self.writer.write_all(b"&lt;"),
-                '>' if !attr_mode => self.writer.write_all(b"&gt;"),
+                '<' => self.writer.write_all(b"&lt;"),
+                '>' => self.writer.write_all(b"&gt;"),
                 c => self.writer.write_fmt(format_args!("{c}")),
             }?;
         }
```

I am willing to submit a PR with the fix and the test case. Thank you!

jdm:
[...]

PR #697, opened 2025-11-30
fix: Escape '<' and '>' when serializing attribute values

pizzacat83:
Fixes https://github.com/servo/html5ever/issues/696.

Applies [the recent change to the HTML Standard](https://github.com/whatwg/html/commit/e21bd3b4a94bfdbc23d863128e0b207be9821a0f) on the attribute serialization logic.


simonwuelker:
Thanks! This does fix the three tests from https://github.com/web-platform-tests/wpt/pull/27501 for servo.

PR #699, opened 2025-12-05
Add iframe `allow` local name

tharkum:
See https://html.spec.whatwg.org/multipage/#attr-iframe-allow


tharkum:
@mrobinson , @jdm < Please review

PR #706, opened 2026-01-09
Prepare web_atoms 0.2.1.

jdm:
#692 introduced a new feature in web_atoms that is used by the crates that depend on it, but that change hasn't been published yet. This blocks publish the new versions of markup5ever, html5ever, and xml5ever.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ rust-version = "1.71.0"
 [workspace.dependencies]
 # Repo dependencies
 tendril = { version = "0.4.3", path = "tendril" }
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
The linked change is technically breaking for anyone who depends on on `web_atoms` with `no-default-features` and who needs serde support. However, that's pretty unlikely given that `web_atoms` didn't previously have any features. And also given that `stylo` and `markup5ever` are the only dependents of `web_atoms` listed on crates.io, and this doesn't affect either of them.

So I think we can probably safely ignore it in this case.

PR #704, opened 2026-01-07
tendril: Add LossyDecoder::new_from_encoding_rs_decoder

simonwuelker:
After merging this change, the new version of `tendril` should be released.

PR #710, opened 2026-01-10
publish versions of html,xml and markup5ever with new tendril release

simonwuelker:
https://github.com/servo/html5ever/pull/704 bumped tendril to 0.5 but there are no versions of markup5ever/html5ever/xml5ever using this release yet, preventing me from using it in servo. I missed this detail in the original PR.

As far as I know it is more or less an unresolved question whether bumping a dependency to a non-backwards-compatible new version should be considered a breaking change, I went with a patch release here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.37.0"
+version = "0.37.1"
 license = "MIT OR Apache-2.0"
 authors = [ "The html5ever Project Developers" ]
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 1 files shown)


atouchet:
> At least `html5ever` and `markup5ever` need to be breaking releases due to #560

Wasn't that included in the 0.37.0 release?

nicoburns:
Ah, I was not aware that we had made a release so recently (version bump was included in with another PR rather than it having a dedicated version bump PR).

yezhizhen:
This version isn't up on [crates.io](https://crates.io/crates/html5ever) yet.

jdm:
Published.

PR #713, opened 2026-01-13
Automatically check semVer compatibility in CI

simonwuelker:
This action will automatically assign the "breaking" or "non-breaking" to PRs based on the output of `cargo-semver-checks`.

Testing: I opened two test PRs in my fork, which were labelled correctly (https://github.com/simonwuelker/html5ever/pulls)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,10 +143,10 @@ jobs:
       - name: Run cargo-semver-checks
         id: semver
         run: |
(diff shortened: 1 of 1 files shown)


simonwuelker:
I've added a line that removes both labels before assigning the new one.

simonwuelker:
Permissions work fine on my fork, so I think the job is (correctly) failing while the job is not in `main`.
I think I need special permission to merge this while the job is failing though?

Issue #716, opened 2026-01-17
The latest version of html5ever (0.37.1) causes a panic facing  `<meta charset="UTF-8">`


niklak:
The latest version of html5ever (0.37.1) causes a panic during document parsing if:

- the document contains `<meta charset="UTF-8">` inside `<head>`, and
- another element inside `<head>` also has a charset attribute (encountered with `link`).

If the `charset` attribute is removed either from `meta`or from `link`, the document is parsed successfully without a **panic**.

### Example
```html
<!DOCTYPE html>
<html>
    <head>
        <title>Test</title>
        <meta charset="UTF-8" />
        <link type="text/css" rel="stylesheet" charset="UTF-8" href="/static/translateelement.css" />
        <script type="text/javascript" charset="UTF-8" src="/static/js/element/main.js"></script>
    </head>
    <body>
        <p>This is a test page contents.</p>
    </body>
</html>
```

jdm:
What's the panic backtrace?

niklak:
backtrace:

```
thread 'parse_doc_meta_charset' panicked at /home/niklak3/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/html5ever-0.37.1/src/driver.rs:133:9:
assertion failed: self.input_buffer.is_empty()
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/panicking.rs:145:5
   3: <html5ever::driver::Parser<Sink> as tendril::stream::TendrilSink<tendril::fmt::UTF8>>::finish
             at /home/niklak3/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/html5ever-0.37.1/src/driver.rs:133:9
```

simonwuelker:
Once again I am wondering if we should just remove the driver module. Servo doesn't use it, and it doesn't simplify things that much.

Thanks for the reproduction!

PR #717, opened 2026-01-17
Handle encoding indicators in the driver module

simonwuelker:
Fixes https://github.com/servo/html5ever/issues/716

### DIFF
--- a/html5ever/src/driver.rs
+++ b/html5ever/src/driver.rs
@@ -116,8 +116,7 @@ where
 impl<Sink: TreeSink> TendrilSink<tendril::fmt::UTF8> for Parser<Sink> {
     fn process(&mut self, t: StrTendril) {
         self.input_buffer.push_back(t);
-        // FIXME: Properly support </script> somehow.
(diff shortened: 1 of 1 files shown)


html5ever/tests/driver.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+impl TreeSink for Sink {
+    type Handle = usize;
+    type Output = Self;
+    type ElemName<'a> = ExpandedName<'a>;
+    fn finish(self) -> Self {
+        self
+    }
```
simonwuelker: This sink implementation is copied from one of the examples

html5ever/src/driver.rs
```diff
@@ -113,11 +113,24 @@ where
     pub input_buffer: BufferQueue,
 }
 
+impl<Sink: TreeSink> Parser<Sink> {
```
mrobinson: There is already a `impl<Sink: TreeSink> Parser<Sink> {` block below, so please merge that with this one.

mrobinson (approved):

Issue #712, opened 2026-01-12
selectedcontent issues


untitaker:
There are a few issues with the selectedcontent implementation of html5ever:

* The implementation of "clone an option into selectedcontent" is wrong. It does not create a new documentfragment but reparents options onto the root of the current document. I'm not sure how that could've happened, I guess that since #560 was opened the spec has been iterated on.
* https://github.com/whatwg/html/commit/4ce63af1 removes the `</option>` transition, and says that the algorithm for "maybe clone an option" needs to be run everytime any option is popped off the stack of open elements. html5ever is lagging behind here, and it seems that runnign something everytime the stack of open elements is truncated requires a larger refactor.

untitaker:
I also think that these regressiontests should be added to html5lib-tests, not sure if that new behavior is covered there.

simonwuelker:
I propose removing https://docs.rs/html5ever/latest/html5ever/interface/trait.TreeSink.html#tymethod.clone_subtree and instead adding https://html.spec.whatwg.org/#clone-an-option-into-a-selectedcontent as a TreeSink callback.

simonwuelker:
I believe the way we track the enabled selectedcontent element of the nearest ancestor select is very wrong - we set the variable here:

https://github.com/servo/html5ever/blob/b4194ce869699dc4a5d89911b9fa87c9e183ce13/html5ever/src/tree_builder/mod.rs#L1408-L1412

and this is the only place where it is mutated. A couple of things that we don't implement:
* The `selectedcontent` element is only relevant when its enabled, which we never check
* It must be the first `selectedcontent` in tree order of its `<select>` ancestor
* the `<select>` must not have a multiple attribute

and there's probably more. The relevant part of the specifcation is  https://html.spec.whatwg.org/#select-enabled-selectedcontent.

PR #719, opened 2026-01-20
Let embedder handle "maybe clone option into selectedcontent" algorithm

simonwuelker:
The current way we track the enabled selectedcontent element of the nearest ancestor select is wrong (see https://github.com/servo/html5ever/issues/712#issuecomment-3770564267). This change moves most of the code for `maybe clone an option into selectedcontent` into `TreeSink`, removing the need to track said element.

Hopefully the amount of code needed to provide a reasonable (yet still incorrect!) implementation of the new method in `rcdom` is enough to convince everyone that we don't want to handle this inside html5ever - it needs very tight integration with the DOM, much more so than anything else we currently have.

Merging this change will also require a new breaking release. It is unfortunate to bump the version again so soon, but the current version has major issues, see https://github.com/servo/html5ever/pull/560#discussion_r2683325031.

Part of https://github.com/servo/html5ever/issues/712
Companion PR for https://github.com/servo/servo/pull/42036

cc @untitaker

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.37.1"
+version = "0.38.0"
 license = "MIT OR Apache-2.0"
 authors = [ "The html5ever Project Developers" ]
 repository = "https://github.com/servo/html5ever"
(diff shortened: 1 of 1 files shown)


simonwuelker:
https://github.com/servo/servo/pull/42036 is approved.

Can someone with release permissions merge this and then publish the new versions?

mrobinson:
I've released the new versions.

PR #723, opened 2026-01-23
Add "alpha" and "colorspace" to web_atoms

simonwuelker:
These are both attribute names relevant in https://html.spec.whatwg.org/multipage/input.html#serialize-a-color-well-control-color, which we need to implement for https://github.com/servo/servo/issues/42117


jdm:
Published.

PR #725, opened 2026-02-02
Fix `Node` drop clobbering children

### LINKED ISSUE #410: rcdom's node's destructor clears all children for nodes that can have strong references outside of the tree
#409 contains this code:
```
            // If this is not the only strong reference, don't drop any children, because
            // apparently somebody else wants them. Just let our reference drop instead.
            if Rc::strong_count(&root) > 1 {
                continue;
            }
```
Our current implementation does not perform this check, but presumably this isn't a case that affects our tests which are largely concerned with the tree structure.

Lucretiel:
`Node::drop` no longer clobbers child nodes that have active external handles. It still takes care to drop iteratively rather than recursively, to avoid an unbounded stack growth. As an added benefit, may be marginally more performant, as it now can avoid `borrow_mut` on the `RefCell`.

Additionally add a regression test for this behavior.

Fixes #410

### DIFF
--- a/rcdom/lib.rs
+++ b/rcdom/lib.rs
@@ -257,28 +257,31 @@ impl Node {
     }
 }
 
+/// Get all of the recursive nodes out of a node and add them to a droplist.
+/// this helps prevent unbounded recursion during drop.
+fn destructure_node(droplist: &mut Vec<Handle>, node: &mut Node) {
+    droplist.extend(mem::take(node.children.get_mut()));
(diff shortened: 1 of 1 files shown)


Lucretiel:
CI failure is some kind of job cancellation.

nicoburns:
Ah, no need to worry about the SemVer check. That isn't working yet and doesn't block merging.

PR #739, opened 2026-05-15
Add a benchmark for serialization and deserialization

mrobinson:
This is the first step toward improvements to HTML parsing and
serialization performance.

### DIFF
--- a/rcdom/benches/deserialize-serialize.rs
+++ b/rcdom/benches/deserialize-serialize.rs
@@ -1,0 +1,92 @@
+#[macro_use]
+extern crate criterion;
+extern crate html5ever;
+
+use std::path::PathBuf;
+use std::{fs, io};
+
+use criterion::{BatchSize, Criterion};
+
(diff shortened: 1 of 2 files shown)


rcdom/benches/deserialize-serialize.rs
```diff
@@ -0,0 +1,101 @@
...
+            },
+            BatchSize::SmallInput,
+        )
+    });
+}
+
+fn deserialize_to_rcdom(stream: &[u8]) -> RcDom {
+    let opts = ParseOpts {
+        tree_builder: TreeBuilderOpts {
+            drop_doctype: true,
```
simonwuelker: why drop the doctype?

simonwuelker (approved):

rcdom/benches/deserialize-serialize.rs
mrobinson: This was a copy and paste error from another one of the benchmarks. I'll remove this.

PR #741, opened 2026-05-17
Remove the ability of the `go!` macro to create enum variants

mrobinson:
This is incredibly confusing when reading the code as it is hard to see
that `go! to Enum Variant1 VariantA` constructs
`Enum(Variant1(VariantA))`. This kind of non-standard Rust makes the
code hard to read by new contributors. In addition, constructing the
variants like this interferes with our ability to make changes using
`rust-analyzer` which is a very useful tool.

This change removes this feature in favor of just writing out the enum
variant directly and starts using the fully qualified name of the state
enum in more places.


simonwuelker (approved):
I would be very happy if we could get rid of `go!` altogether

PR #742, opened 2026-05-18
Remove the `pop_except_from!` macro

mrobinson:
This the first step toward a more generic SIMD optimized path for
scanning ahead and also for a potentially easier way of bailing out of
this loop using the `?` Rust operator.


simonwuelker (approved):
> and also for a potentially easier way of bailing out of
this loop using the ? Rust operator.

The unstable `Try` trait would be very helpful here: https://github.com/rust-lang/rust/issues/84277

PR #745, opened 2026-06-16
feat: add byte tracking to html5ever

### LINKED ISSUE #734: Byte-accurate source positions in TreeSink
> Relates to [#48](https://github.com/servo/html5ever/issues/48) and [#492](https://github.com/servo/html5ever/issues/492).

[Cadmus](https://github.com/OGKevin/cadmus) is a Reader software for e-readers. 

The original HTML parsing implementation was hand written, and is not spec complete and contains some minor bugs. 
While working on https://github.com/OGKevin/cadmus/issues/343, I decided to give html5ever a try to replace at least the parsing bit.

It turns out, that html5ever does not provide a way to accurately know a position within a document.
The reason this is needed:

- Save and restore reading positions across sessions
- Persist bookmarks and annotations
- Resolve `#anchor-id` URI fragment links

For all of these to work correctly across re-parses, the offset stored on each node
must be the **byte position of that node's opening token in the source string**. It
needs to be stable and comparable to the raw byte sizes of the EPUB spine entries.

For the rendering of dictionary HTML, html5ever covers the use case, as there is no need for position tracking,
but it can't be used for EPUB rendering.

Would it be interesting to add the ability for when parsing a document, to store byte offsets, as this
would be the most stable way to refer to a position within a document and doesn't matter
which parsing system is being used.

OGKevin:
# Byte Tracking

This PR introduced the ability for `TreeSink` implementations to know byte positions of Tokens. This is done by making `BufferQueue` track byte consumption and expose methods for `Tokenizer` to get this info, which than passes it forwards to `TreeSink`.

## BufferQueue

`BufferQueue` gets a `bytes_consumed` field that tracks bytes manipulated via:

- `pop_except_from`
- `advance_bytes_consumed` 
- `retreat_bytes_consumed`
- `eat`
- `next`

The "consuming" methods (pop, eat, next) automatically _advance_ the counter. While `push_back` and `push_front`  don't move the counter. This is due to some code paths pushing front without actually consuming bytes, so this would corrupt the counter. Hence it's the callers responsibility to retreat.

## Tokenizer

The main use case of this PR is to know _where_ a token began in bytes, not necessarily the raw byte position of a character (what `BufferQueue::bytes_consumed` tracks. 

Since Tokenizer is aware of "Tokens" e.g. Tag, Comment and Character tokens, it is in a good position to dictate what the correct byte is that needs to be reported to `TreeSink`.

For Tag, Comment and Doctype, the byte offset is that start of the token. So for tag  `<some-tag></some-tag>`, its offset would be `0`. And for tag `p` in `<some-tag><p></p></some-tag>` it would be `10`.

For character tokens, it wold be that start of the first character.
e.g. `<p>abc</p> the text `abc` would have offset `3`.

This is done by keeping track of when we saw the last `<` when processing a token. This is stored as `token_start_byte`. As we continue to process, we always write to `last_token_end_byte` at the end of each token processing so it can be used when we get `CharacterTokens`. In the above example, it would contain the byte of `a`. As when we get the character token `abc` we would be at byte `6` but we want to communicate that the character token started at `3`.

## TokenSink

Token sink get's a new 

```rust
    #[cfg(feature = "source-positions")]
    fn set_current_byte(&self, _byte_offset: u64) {}
```

Method instead of updating

```rust
    /// Process a token.
    fn process_token(&self, token: Token, line_number: u64) -> TokenSinkResult<Self::Handle>;
```

And breaking the signature.

## TreeSink

For folks who implement their own Dom struct, `TreeSink` is able to also pass the byte offset forward. 

An example of this can be seen in `source-positions-integration.rs`

```rust
    impl TreeSink for ByteCapturingDOM {
```

## Conclusion

`BufferQueue` holds raw position, Tokenizer converts raw position to an offset, and this gets passed down to TokenSink and TreeSink.

<sub>This was manually written ⌨️. Mainly to proof that human 🧠 and hands were used . </sub>

---

### Cargo

I had to make some small changes to the `cargo.yaml` files so that cargo would be happy to compile this project as a submodule: https://github.com/OGKevin/cadmus/blob/65cdaaabb5ab795c668b32a4206f8a49fc9e2661/Cargo.toml#L26-L28

### `xhtml-self-closing` feature

This feature is needed to properly process self closing tags such as `<title/>`, which might not be valid html5, but seems to be valid xhtml. Without this feature, html5ever swallows everything while looking for the closing title tag.

`xhtml-self-closing-integration.rs` has test case that showcases this. Well, not the "bug" per se, just how this feature fixes the issue. 

Ref: https://github.com/baskerville/plato/issues/426#issuecomment-4690477643

---

🤖 ~~was used to generate some of the rustdocs and tests.~~ Was also used to educate me on the project and brainstorm implementation path. 

---

Closes: https://github.com/servo/html5ever/issues/734

Ref:
- https://github.com/servo/html5ever/issues/734
- https://github.com/OGKevin/cadmus/issues/384
- https://github.com/OGKevin/cadmus/pull/599


simonwuelker:
> 🤖 was used to generate some of the rustdocs and tests. Was also used to educate me on the project and brainstorm implementation path.

Note that servo does not accept AI-generated contributions: https://book.servo.org/contributing/getting-started.html#ai-contributions.

Admittedly we should do a better job of communicating that this applies to all repositories under the servo org, not just `servo/servo`.

OGKevin:
Ahh, that might make things interesting now. I wasn't aware, mybad. 

What now? As I see 2 solutions:

1. Remove the tests and rustdocs, as thats the _easiest_ way to remove all 🤖 touched code, but would also significantly lower the quality of the PR.
2. Re-write them, but I'm now biased to what I've seen, reviewed and agreed with, so my manual rewrite might not be _significantly_ different. Unless you have some feedback on it that makes it look diff. However, how would I then proof that was scaffolded by 🤖 is now 100% manually typed? 

The tests turned out to be useful as they caught some gaps in the implementation, but since they've "served their purpose" I'm fine with removing them. As In this specific situation, I've similar tests in Cadmus that covers similar cases.

Do note that tests are the biggest diff: 
image


---

I'm not challenging the usage of 🤖, just not sure how to move forward. Unfortunately, my daily life has come to 🤖 code reviewer. So I take the small manual implementations where I can, was just lazy when it came to tests and rustdocs 🙉 😭.

simonwuelker:
I'd be fine with you rewriting the tests and rustdoc.

> but I'm now biased to what I've seen, reviewed and agreed with, so my manual rewrite might not be significantly different.

That's okay.

>  However, how would I then proof that was scaffolded by 🤖 is now 100% manually typed?

You don't need to prove that. We rely on the honesty of contributors when rejecting AI generated patches (unless the patch is obviously slop). Thanks for being upfront about it.

OGKevin:
I think this is ready for review 🤞🏾. I believe I've touched up on all the parts that were fully generated.

So far, I haven't encountered any bugs 🙏🏾, but would like to keep this project as a submodule, so I always have the ability to make changes if needed, since its now an integral part of Cadmus. That being said:

The only thing I'm unsure about are the changes in cargo.toml. Not sure if there is a way to add this project as a submodule. I've read https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html 2 times now and the only solution that seems to work is by updating cargo.toml in this project.

I was hoping maybe this could work:

```toml
[patch.crates-io]
html5ever  = { path = "thirdparty/html5ever", package = "html5ever"}
```

But it doesn't. 

```
error: failed to load source for dependency `html5ever`

Caused by:
  unable to update /home/kevin/code/github.com/ogkevin/cadmus/thirdparty/html5ever

Caused by:
  found a virtual manifest at `/home/kevin/code/github.com/ogkevin/cadmus/thirdparty/html5ever/Cargo.toml` instead of a package manifest
```

```
thirdparty/html5ever/
├── AUTHORS
├── Cargo.lock
├── Cargo.toml
├── COPYRIGHT
├── html5ever
├── LICENSE-APACHE
├── LICENSE-MIT
├── markup5ever
├── rcdom
├── README.md
├── RELEASING.MD
├── rustfmt.toml
├── target
├── tendril
├── web_atoms
└── xml5ever

8 directories, 9 files
```

Not sure if there is something I'm missing tho 🙈.

html5ever/src/tokenizer/interface.rs
```diff
@@ -130,6 +130,13 @@ pub trait TokenSink {
     /// Signal that tokenization reached the end of the document.
     fn end(&self) {}
 
+    /// Called just before each token is dispatched to [`process_token`],
+    /// with the number of UTF-8 bytes consumed from the input so far.
+    ///
+    /// The default implementation is a no-op.
+    #[cfg(feature = "source-positions")]
+    fn set_current_byte(&self, _byte_offset: u64) {}
```
simonwuelker: I would strongly prefer passing the byte position as an offset to `process_token` instead of adding a new method that will be called right before.

You can add something like a `struct SourcePosition` which would contain the line number and (optionally) the byte offset.

html5ever/src/tokenizer/mod.rs
```diff
@@ -621,7 +682,20 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
...
+            #[cfg(feature = "source-positions")]
+            {
+                let c = input.next();
+                if let Some(c) = c {
+                    let pos = input.bytes_consumed();
+                    self.current_byte.set(pos);
+                    if c == '<' {
+                        self.token_start_byte
+                            .set(pos.saturating_sub(c.len_utf8() as u64));
+                    }
```
simonwuelker: This doesn't work - not every `<` indicates the start of a token (for example in the PLAINTEXT state [^1] it is simply emitted as any other character).

[^1]: https://html.spec.whatwg.org/#plaintext-state

html5ever/src/tokenizer/mod.rs
```diff
@@ -757,6 +831,20 @@ impl<Sink: TokenSink> Tokenizer<Sink> {
                         // This CPU is guaranteed to support SIMD due to the is_supported_simd_feature_detected check above
                         let result = unsafe { self.data_state_simd_fast_path(&mut front_buffer) };
 
+                        #[cfg(feature = "source-positions")]
+                        if let Some(ref r) = result {
+                            let n = match r {
+                                SetResult::NotFromSet(ref t) => t.len() as u64,
```
simonwuelker: no one-letter variable names please

html5ever/src/tokenizer/mod.rs
```diff
@@ -2379,3 +2469,236 @@ mod test {
         assert_eq!(results, expected);
     }
 }
+
+#[cfg(all(test, feature = "source-positions"))]
+mod test_source_positions {
```
simonwuelker: integration tests should live in `html5ever/tests`,

html5ever/Cargo.toml
```diff
@@ -5,25 +5,31 @@ documentation = "https://docs.rs/html5ever"
...
-authors.workspace = true
-repository.workspace = true
-edition.workspace = true
-rust-version.workspace = true
+version = "0.39.0"
+license = "MIT OR Apache-2.0"
+authors = [ "The html5ever Project Developers" ]
+repository = "https://github.com/servo/html5ever"
+edition = "2021"
+rust-version = "1.71.0"
```
simonwuelker: You'll need to bump the version if we publish this, but don't modify the other fields please.

markup5ever/util/buffer_queue.rs
```diff
@@ -51,6 +53,12 @@ pub enum SetResult {
 pub struct BufferQueue {
     /// Buffers to process.
     buffers: RefCell<VecDeque<StrTendril>>,
+    /// Total number of UTF-8 bytes consumed from this queue so far.
+    ///
+    /// Used by the tokenizer to surface byte-accurate source offsets via
+    /// [`TokenSink::set_current_byte`] and [`TreeSink::set_current_byte`].
+    #[cfg(feature = "source-positions")]
+    bytes_consumed: Cell<u64>,
```
simonwuelker: I don't understand why we need to count the consumed bytes here. A `BufferQueue` just stores input bytes - it is not tied to the tokenizer in any way. Any book-keeping of the consumed input should happen inside the tokenizer.

simonwuelker: The number of bytes should be a `usize`. That avoids some casting back and forth in the rest of the code.

simonwuelker (changes requested):
These are a couple things I found while skimming through the code. The byte-tracking looks pretty invasive, which concerns me. Ideally we would keep the number of `#[cfg(feature = "source-positions")]` low, much lower than it currently is. I don't think any changes to the internals of `BufferQueue` should be necessary.

> I had to make some small changes to the cargo.yaml files so that cargo would be happy to compile this project as a submodule: https://github.com/OGKevin/cadmus/blob/65cdaaabb5ab795c668b32a4206f8a49fc9e2661/Cargo.toml#L26-L28

That is not how we intend for html5ever to be consumed. Please don't change our workspace configuration.

> I was hoping maybe this could work:

try `html5ever  = { path = "thirdparty/html5ever/html5ever" }`, that works for me when developing servo.

> xhtml-self-closing feature
This feature is needed to properly process self closing tags such as <title/>, which might not be valid html5, but seems to be valid xhtml. Without this feature, html5ever swallows everything while looking for the closing title tag.
xhtml-self-closing-integration.rs has test case that showcases this. Well, not the "bug" per se, just how this feature fixes the issue.

That is completely unrelated to this PR. Please open a different one for this bug.

nicoburns:
> This feature is needed to properly process self closing tags such as <title/>, which might not be valid html5, but seems to be valid xhtml. Without this feature, html5ever swallows everything while looking for the closing title tag.

This is what the `xml5ever` crate is for. It will parse XHTML documents correctly.

OGKevin:
> try html5ever  = { path = "thirdparty/html5ever/html5ever" }, that works for me when developing servo.

This is what I have! And without the changes, I get the same
> failed to load source for dependency `html5ever'

error 😔. I'll spend some more time on it, worst case I'll just keep the workspace config changes in the fork on a separate commit 👍🏾 Since 
> That is not how we intend for html5ever to be consumed.

is valid 🤝 

---

> This is what the xml5ever crate is for. It will parse XHTML documents correctly.

Ahh I missed this, if thats the case, I should be using xml5ever instead 🤝. I was under the impression it was just for `xml` 😭. Nevertheless, I will still attempt at getting this PR merged for `html5ever`. Afterwards, I'll replicate the changes to `xml5ever`. 

In this case, the `xhtml-self-closing` can be removed entirely. I'm not certain if it should still be classified as a "bug" @simonwuelker? As this is a clear case of user error(xml5ever for xhtml instead of html5ever) + invalid spec (xhtml vs html) 🤔.

markup5ever/util/buffer_queue.rs
OGKevin: My thinking is that since the buffer queue is in charge of the raw input, it should be the source of truth to what has been consumed or not? 

In theory, something could happen in buffer queue that the tokenizer is not aware of and corrupt the counter 🤔

html5ever/src/tokenizer/mod.rs
OGKevin: For this specific implementation, this is not a problem. As the value is only _read_ after it has been determined that we're processing a non character token:

```rust
Token::TagToken(_) | Token::CommentToken(_) | Token::DoctypeToken(_) => {
                    self.token_start_byte.get()
                },
```

This is not under test however, I'll add a test case later something like:

```rust
fn check_byte_offsets_text_content_with_token_start() {
        let entries = tokenize_bytes("<p>hello<world</p>");
```

Hopefully it proofs me right 😆.

html5ever/src/tokenizer/mod.rs
OGKevin: This is not integration testing actually, its similar to the main `mod test` similar to 

```rust
// LinesMatch implements the TokenSink trait. It is used for testing to see
    // if current_line is being updated when process_token is called. The lines
    // vector is a collection of the line numbers that each token is on.
    struct LinesMatch {
        tokens: RefCell<Vec<Token>>,
        current_str: RefCell<StrTendril>,
        lines: RefCell<Vec<(Token, u64)>>,
    }
```

but I wanted to separate it. Would it be okay to do:

```rust

mod test {
   // existing

   #[cfg(feature = "source-positions")]
   mod test_source_positions {
    // tests
   }
}
```

?

simonwuelker:
> This is what I have!

Note the extra `/htmlever` in the path compared to what you posted (:

> In this case, the xhtml-self-closing can be removed entirely. I'm not certain if it should still be classified as a "bug" @simonwuelker? As this is a clear case of user error(xml5ever for xhtml instead of html5ever) + invalid spec (xhtml vs html) 🤔.

If xml5ever has the correct behaviour then this is not a bug, no.

OGKevin:
> Note the extra /htmlever in the path compared to what you posted (:

🙈 

> I had to make some small changes to the cargo.yaml files so that cargo would be happy to compile this project as a submodule: OGKevin/cadmus@65cdaaa/Cargo.toml#L26-L28

That link, and what has been comitted is: 

```rust 
[patch.crates-io]
html5ever  = { path = "thirdparty/html5ever/html5ever" }
markup5ever = { path = "thirdparty/html5ever/markup5ever" }
```

Which is the same?

> I was hoping maybe this could work:

this is just another variation that _completely_ doesn't work. 
And the above `"thirdparty/html5ever/html5ever` also doesn't work without the changes in cargo.toml in html5ever 😔.
As it doesn't like that cargo.toml in `html5ever` is defering fields to workspace:

```bash
$ cargo check
error: failed to load source for dependency `html5ever`

Caused by:
  unable to update /Users/x/code/github.com/ogkevin/cadmus/thirdparty/html5ever/html5ever

Caused by:
  failed to parse manifest at `/Users/x/code/github.com/ogkevin/cadmus/thirdparty/html5ever/html5ever/Cargo.toml`

Caused by:
  error inheriting `edition` from workspace root manifest's `workspace.package.edition`

Caused by:
  `workspace.package.edition` was not defined
```

Despite it being defined:

```
[workspace.package]
version = "0.39.0"
license = "MIT OR Apache-2.0"
authors = [ "The html5ever Project Developers" ]
repository = "https://github.com/servo/html5ever"
edition = "2021"
rust-version = "1.71.0"
```

I'll spend some time on it again later today, and if not I'll just keep it separate 👍🏾

nicoburns:
If you can, try using `path = '...'` in the regular dependency section rather than using a patch.

OGKevin:
> If you can, try using path = '...' in the regular dependency section rather than using a patch.

```
❯ jj diff --git "**/Cargo.toml"
diff --git a/Cargo.toml b/Cargo.toml
index 329d1e8f5e..c20ab84e97 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,9 +23,9 @@
 [workspace.lints]
 rustdoc.private_intra_doc_links = "allow"

-[patch.crates-io]
-html5ever  = { path = "thirdparty/html5ever/html5ever" }
-markup5ever = { path = "thirdparty/html5ever/markup5ever" }
+# [patch.crates-io]
+# html5ever  = { path = "thirdparty/html5ever/html5ever" }
+# markup5ever = { path = "thirdparty/html5ever/markup5ever" }

 [profile.release-minsized]
 inherits = "release"
diff --git a/crates/core/Cargo.toml b/crates/core/Cargo.toml
index 4a381182d2..35f0624a36 100644
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -41,7 +41,7 @@
 walkdir = "2.5.0"
 globset = "0.4.18"
 fxhash = "0.2.1"
-html5ever = { version = "0.39", features = ["source-positions", "xhtml-self-closing"] }
+html5ever = {path = "../../thirdparty/html5ever/html5ever", version = "0.39", features = ["source-positions", "xhtml-self-closing"] }
 roxmltree = "0.21"
 humanize-bytes = "1.0.6"
 humantime = "2"

plato [$!?] via  v24.14.1 via 🦀 v1.96.0 via ❄️  impure (devenv-shell-env) 🥋 mvzm | (no description set) as 👩‍💻
❯ cargo check
error: failed to load manifest for workspace member `/home/kevin/code/github.com/ogkevin/plato/crates/core`
referenced by workspace at `/home/kevin/code/github.com/ogkevin/plato/Cargo.toml`

Caused by:
  failed to load manifest for dependency `html5ever`

Caused by:
  failed to parse manifest at `/home/kevin/code/github.com/ogkevin/plato/thirdparty/html5ever/html5ever/Cargo.toml`

Caused by:
  error inheriting `edition` from workspace root manifest's `workspace.package.edition`

Caused by:
  `workspace.package.edition` was not defined
```

---

upon a bit further investigation, defining these fields in https://github.com/OGKevin/cadmus/blob/a72e31d1e51a50fb45403527b7de553ebb936d9c/Cargo.toml makes the error go away for that field as well 🤯. 

```
❯ jj diff --git "**/Cargo.toml"
diff --git a/Cargo.toml b/Cargo.toml
index 329d1e8f5e..78cb9277f9 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@
   "xtask",
 ]

+[workspace.package]
+edition = "2024"
+
 [workspace.dependencies]
 rustls = { version = "0.23", default-features = false, features = [
   "ring",
@@ -19,13 +22,15 @@
 ] }
 cc = "1"
 rust-embed = "8"
+log = "0.4"
+markup5ever = { path = "thirdparty/html5ever/markup5ever" }

 [workspace.lints]
 rustdoc.private_intra_doc_links = "allow"

-[patch.crates-io]
-html5ever  = { path = "thirdparty/html5ever/html5ever" }
-markup5ever = { path = "thirdparty/html5ever/markup5ever" }
+# [patch.crates-io]
+# html5ever  = { path = "thirdparty/html5ever/html5ever" }
+# markup5ever = { path = "thirdparty/html5ever/markup5ever" }

 [profile.release-minsized]
 inherits = "release"
diff --git a/crates/core/Cargo.toml b/crates/core/Cargo.toml
index 4a381182d2..35f0624a36 100644
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -41,7 +41,7 @@
 walkdir = "2.5.0"
 globset = "0.4.18"
 fxhash = "0.2.1"
-html5ever = { version = "0.39", features = ["source-positions", "xhtml-self-closing"] }
+html5ever = {path = "../../thirdparty/html5ever/html5ever", version = "0.39", features = ["source-positions", "xhtml-self-closing"] }
 roxmltree = "0.21"
 humanize-bytes = "1.0.6"
 humantime = "2"

plato [$!?] via  v24.14.1 via 🦀 v1.96.0 via ❄️  impure (devenv-shell-env) 🥋 mvzm | (no description set) as 👩‍💻
❯ cargo check
error: failed to load manifest for workspace member `/home/kevin/code/github.com/ogkevin/plato/crates/core`
referenced by workspace at `/home/kevin/code/github.com/ogkevin/plato/Cargo.toml`

Caused by:
  failed to load manifest for dependency `html5ever`

Caused by:
  failed to parse manifest at `/home/kevin/code/github.com/ogkevin/plato/thirdparty/html5ever/html5ever/Cargo.toml`

Caused by:
  error inheriting `rust-version` from workspace root manifest's `workspace.package.rust-version`

Caused by:
  `workspace.package.rust-version` was not defined
```

But at some point it wants me to also list all the dependencies:

```
❯ cargo check
error: failed to load manifest for workspace member `/home/kevin/code/github.com/ogkevin/plato/crates/core`
referenced by workspace at `/home/kevin/code/github.com/ogkevin/plato/Cargo.toml`

Caused by:
  failed to load manifest for dependency `html5ever`

Caused by:
  failed to parse manifest at `/home/kevin/code/github.com/ogkevin/plato/thirdparty/html5ever/html5ever/Cargo.toml`

Caused by:
  error inheriting `memchr` from workspace root manifest's `workspace.dependencies.memchr`

Caused by:
  `dependency.memchr` was not found in `workspace.dependencies`
  ```

So I'll just cherry pick the cargo.toml file changes in a stacked commit and exclude them from this MR 👍🏾

File: html5ever/src/tokenizer/mod.rs (hunk 2 above)
Comment (OGKevin) [reply html5ever/src/tokenizer/mod.rs]: Hacky test proofs it indeed: 

```rust   
     impl TokenSink for PlaintextAwareSink {
        type Handle = ();

        fn process_token(&self, token: Token, _line_number: u64) -> TokenSinkResult<Self::Handle> {
            let is_plaintext = matches!(
                token,
                TagToken(Tag {
                    kind: StartTag,
                    ref name,
                    ..
                }) if name == &LocalName::from("plaintext")
            );
            let res = self.inner.process_token(token, _line_number);
            if is_plaintext {
                TokenSinkResult::Plaintext
            } else {
                res
            }
        }
```

with 

```rust
  buf.push_back(StrTendril::from("<plaintext><p>hello<world</p>"));
    let _ = tok.feed(&buf);
    tok.end();
    let entries = tok.sink.inner.entries.take();

    assert_eq!(
        entries,
        vec![
            (start("plaintext"), 0),
            (chars("<p>hello<world</p>"), 11),
        ]
    );
```

```bash
❯ cargo-nextest ntr --features source-positions check_byte_offsets_text_content_with_start_token
   Compiling html5ever v0.39.0 (/home/kevin/code/github.com/ogkevin/plato/thirdparty/html5ever/html5ever)
   Compiling markup5ever_rcdom v0.39.0 (/home/kevin/code/github.com/ogkevin/plato/thirdparty/html5ever/rcdom)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.29s
────────────
 Nextest run ID b41f2346-e90a-4e3a-a704-787fd588bc46 with nextest profile: default
    Starting 1 test across 17 binaries (29419 tests skipped)
        PASS [   0.005s] html5ever tokenizer::test_source_positions::check_byte_offsets_text_content_with_start_token
────────────
     Summary [   0.036s] 1 test run: 1 passed, 29419 skipped
```

Would you like for this to be under test? And importantly, did I understand you correctly and does this cover what you meant 👀

OGKevin:
@simonwuelker can you have a look at the replies on your review comments 👀 so I can make the required changes as well re reverting buffer queue, split off test, `<` etc.

markup5ever/util/buffer_queue.rs
simonwuelker: The tokenizer is the one that ultimately consumes the input and its also where we count newlines already. Please do move the byte counting to there. Nothing prevents a user from switching between different buffer queues.

OGKevin: > Nothing prevents a user from switching between different buffer queues.

didn't think of this 🧠 🤝, addressed.

html5ever/src/tokenizer/mod.rs
OGKevin: What shall be done here? Cover it with test or think of a diff implementation? In theory this should just work.

OGKevin:
@simonwuelker what needs to be done to get this merged 👀 

A follow up PR is needed to implement this in `xml5ever` as well. But rather have this one merged / approved and afterwards "copy paste" the implementation there.

simonwuelker:
Sorry for being unresponsive on this, I intend to take another look later this week.

PR #750, opened 2026-06-30
Update phf to 0.14. Bump versions.

xu-cheng:
Also bump MSRV to 1.85 as required by phf.

Per RELEASE.md, phf updates are breaking changes. Thus, bump versions for both web_atom and html5ever.

Depends on https://github.com/servo/string-cache/pull/298

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Install stable toolchain
         run: |
           rustup set profile minimal
-          rustup override set 1.71.0
+          rustup override set 1.85.0
 
       - run: cargo check --lib --all-features
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,30 +10,30 @@ members = [
 ]
 
 [workspace.package]
-version = "0.39.0"
+version = "0.40.0"
 license = "MIT OR Apache-2.0"
 authors = [ "The html5ever Project Developers" ]
 repository = "https://github.com/servo/html5ever"
 edition = "2021"
-rust-version = "1.71.0"
+rust-version = "1.85"
 
 [workspace.dependencies]
 # Repo dependencies
 tendril = { version = "0.5", path = "tendril" }
-web_atoms = { version = "0.2.5", path = "web_atoms" }
-markup5ever = { version = "0.39", path = "markup5ever" }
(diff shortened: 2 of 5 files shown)


xu-cheng:
Hi, any chance to merge this and publish a new version of html5ever? Thanks.

nicoburns (approved):

nicoburns:
@mrobinson @Loirooriol @jdm @simonwuelker How urgent are the security fixes that seem to open against this repo recently?

This PR looks good to me, but I think landing it may mean that we can't upgrade the html5ever version in Servo until such time as `cssparser` and Stylo are also upgraded to phf 0.14 (Stylo is easy enough as we have our own fork, but `cssparser` will require coordination). So I'm not sure if we want to hold off on that and put out a patch release with the fixes first?

Loirooriol:
Will this upgrade require change changes in cssparser? Or what's the coordination that you are concerned about?

nicoburns:
> Will this upgrade require change changes in cssparser? Or what's the coordination that you are concerned about?

`cssparser` and `stylo`/`selectors` also need an `phf` upgrade. If we want only one version of phf in Servo, then those upgrades all need to happen in sync. So the concern is that if we land this and the `phf` upgrade in `cssparser` / Stylo is slow then that might block us upgrading html5ever in Servo until that's resolved.

simonwuelker:
> How urgent are the security fixes that seem to open against this repo recently?

* https://github.com/servo/html5ever/issues/753: Is pretty bad.
* https://github.com/servo/html5ever/issues/752: Should not cause problems for any reasonable implementation of `Read`, but is still a source of potential UB. Looks like `read_to_tendril` is mostly used in tests, benchmarks and example code for `tendril`.
* https://github.com/servo/html5ever/issues/751: No idea, depends on whether or not `Vec::with_capacity` actually over-allocates in practice.


I would strongly prefer not merging changes that prevent us from using the latest version of html5ever in servo.

SimonSapin:
If it comes to that I think we should accept temporarily having two versions of phf in Servo rather than block security fixes

Still, let’s track moving Servo’s dependency graph to phf 0.14: https://github.com/servo/servo/issues/47431

SimonSapin:
By the way @nicoburns if anything Servo-related is blocked on reviews or releases in cssparser feel free to ping me

PR #755, opened 2026-07-03
Fix data race in SendTendril by removing Clone

### LINKED ISSUE #753: Remove  #[derive(Clone)]  from  SendTendril  to prevent data races
`SendTendril` exists to safely transfer a `Tendril<F, NonAtomic>` across threads. But it derives `Clone`, which just delegates to the underlying non-atomic clone implementation.

Because `SendTendril` is `Send`, you can clone it and send the copies to different threads. When those threads drop or modify the tendrils, they end up concurrently modifying a non-atomic `Cell<usize>` reference count. This triggers a data race (UB) and leads to memory corruption like double-frees.

**Fix:**
    Drop `#[derive(Clone)]` from `SendTendril`.

matteoldani:
Removed  `#[derive(Clone)]`  from  `SendTendril` . This leverages the compiler to strictly enforce its intended invariant of unique ownership, making it impossible to alias the non-atomic buffer across threads.

  Resolves #753

### DIFF
--- a/tendril/src/tendril.rs
+++ b/tendril/src/tendril.rs
@@ -1228,7 +1228,6 @@ where
 ///
 /// A `SendTendril` may be produced by `Tendril.into_send()` or `SendTendril::from(tendril)`,
 /// and may be returned to a `Tendril` by `Tendril::from(self)`.
-#[derive(Clone)]
 pub struct SendTendril<F>
 where
     F: fmt::Format,


simonwuelker (approved):
Huh, I thought this would cause problems for servo but apparently its not used at all. Well, better get rid of it quickly then. Thanks for the fix!

PR #756, opened 2026-07-03
Fix uninitialized memory exposure in IO reads

### LINKED ISSUE #752: Zero-initialize buffers in  read_to_tendril  and  TendrilSink::read_from
Currently, `ReadExt::read_to_tendril` and `TendrilSink::read_from` both expand their buffers via `push_uninitialized()` before passing the slice to a generic `io::Read` trait object.

There's actually already a `FIXME` comment in the code acknowledging this:
    `// FIXME: this exposes uninitialized bytes to a generic R type`
    `// this is fine for R=File which never reads these bytes,`
    `// but user-defined types might.`
    `// The standard library pushes zeros to Vec<u8> for that reason.`

Passing uninitialized heap memory to a generic reader is UB and can leak leftover heap data if a custom reader misbehaves and inspects the buffer.

 **Fix:**
Just like the standard library does for `Vec`, we should zero-initialize the memory using `std::ptr::write_bytes` before passing it to `read()`. There might be a performance hit, I'm happy to benchmark it to check for any potential regression if needed.

matteoldani:
This PR addresses an API soundness bug where uninitialized heap memory could be exposed to safe  `io::Read`  implementations.

  Currently, both  `ReadExt::read_to_tendril`  and  `TendrilSink::read_from`  expand their internal buffers via  `push_uninitialized()`  before passing the resulting slice to a generic  `io::Read`  trait object. As noted by an existing  FIXME  comment in the codebase:
```
  │  // FIXME: this exposes uninitialized bytes to a generic R type
  │  // this is fine for R=File which never reads these bytes,
  │  // but user-defined types might.
```
  While standard readers like  std::fs::File  only write to the buffer, a poorly written or unconventional custom  io::Read  implementation might attempt to read from the provided slice. Because these tendril functions are safe to call, exposing
  uninitialized memory this way technically violates Rust's strict memory safety guarantees and triggers Undefined Behavior under Miri.

  Changes:

  • Following the precedent of the standard library (e.g.,  `Vec` ), this PR explicitly zero-initializes the newly appended buffer slice using  `std::ptr::write_bytes`  before passing it to the generic reader.
  • Using  `write_bytes`  avoids the UB of creating safe Rust references ( `&mut [u8] `) to uninitialized memory. It also compiles down directly to a highly optimized LLVM  `memset`  intrinsic, ensuring there is no meaningful performance regression.

Resolves #752

### DIFF
--- a/tendril/src/stream.rs
+++ b/tendril/src/stream.rs
@@ -77,20 +77,14 @@ where
     {
         const BUFFER_SIZE: u32 = 4 * 1024;
         loop {
-            let mut tendril = Tendril::<F, A>::new();
-            // FIXME: this exposes uninitialized bytes to a generic R type
-            // this is fine for R=File which never reads these bytes,
-            // but user-defined types might.
-            // The standard library pushes zeros to `Vec<u8>` for that reason.
-            unsafe {
-                tendril.push_uninitialized(BUFFER_SIZE);
-            }
+            let mut tendril = Tendril::<fmt::Bytes, A>::new();
+            tendril.extend_with_byte(BUFFER_SIZE, 0);
             loop {
                 match r.read(&mut tendril) {
                     Ok(0) => return Ok(self.finish()),
                     Ok(n) => {
                         tendril.pop_back(BUFFER_SIZE - n as u32);
-                        self.process(tendril);
+                        self.process(unsafe { tendril.reinterpret_without_validating() });
                         break;
                     },
                     Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {},
--- a/tendril/src/tendril.rs
+++ b/tendril/src/tendril.rs
@@ -1393,13 +1393,7 @@ where
                 if new_write_size < DEFAULT_BUF_SIZE {
                     new_write_size *= 2;
                 }
-                // FIXME: this exposes uninitialized bytes to a generic R type
-                // this is fine for R=File which never reads these bytes,
-                // but user-defined types might.
-                // The standard library pushes zeros to `Vec<u8>` for that reason.
-                unsafe {
-                    buf.push_uninitialized(new_write_size);
-                }
+                buf.extend_with_byte(new_write_size, 0);
             }
 
             match self.read(&mut buf[len..]) {
@@ -1441,6 +1435,24 @@ where
     #[inline(always)]
     fn flush(&mut self) -> io::Result<()> {
         Ok(())
+    }
+}
+
+impl<A> Tendril<fmt::Bytes, A>
+where
+    A: Atomicity,
+{
+    /// Extend the tendril with `n` copies of a given byte.
+    ///
+    /// This grows the tendril and initializes the new area with `byte`.
+    #[inline]
+    pub fn extend_with_byte(&mut self, n: u32, byte: u8) {
+        unsafe {
+            let start = self.len32();
+            self.push_uninitialized(n);
+            let ptr = self[start as usize..].as_mut_ptr();
+            std::ptr::write_bytes(ptr, byte, n as usize);
+        }
     }
 }
 


tendril/src/tendril.rs
```diff
@@ -1399,6 +1399,8 @@ where
                 // The standard library pushes zeros to `Vec<u8>` for that reason.
                 unsafe {
                     buf.push_uninitialized(new_write_size);
+                    let ptr = buf[len..].as_mut_ptr();
+                    std::ptr::write_bytes(ptr, 0, new_write_size as usize);
                 }
```
simonwuelker: Given that the buffer is not actually uninitialized anymore I think we should be able to get rid of the `unsafe`. This might require adding a new API, something akin to [`Vec::resize`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize).

tendril/src/tendril.rs
simonwuelker: We can get rid of these FIXME comments now.

matteoldani: Yep, leftover, sorry. I'll remove them (same in stream.rs).

tendril/src/tendril.rs
matteoldani: True. 

I'll push a change that introduces `push_zeros(n)` that imho aligns with the overall tendril approach of "push_<>" rather then the absolute resize.

tendril/src/tendril.rs
```diff
@@ -1464,6 +1458,19 @@ where
             self.set_len(new_len);
         }
     }
+
+    /// Push zeroed bytes onto the end.
+    ///
+    /// This grows the tendril and initializes the new area with 0s.
+    #[inline]
+    pub fn push_zeros(&mut self, n: u32) {
```
matteoldani: FYI: @simonwuelker this is the new API

simonwuelker: Two things that I forgot during the first review:

* We should only expose this for `Tendril<fmt::Bytes, A>`, not for all formats, as we might violate the internal invariants of the tendril otherwise.
* Let's pass the byte that should be written as an argument, that way the method is actually useful in a more general context. Maybe call the method `extend`?

Apologies for not thinking of this earlier :/

matteoldani: No worries, I will work on them tomorrow. 

Regarding the API shape, I don't have a strong preference. As I mentioned in the previous comment I kept the "push" shape since it seemed more common in tendril API but happy to refactor it into an `extend`

matteoldani: Sorry for the delay, I've pushed the new patch.

tendril/src/stream.rs
```diff
@@ -77,20 +77,14 @@ where
...
-            }
+            let mut tendril = Tendril::<fmt::Bytes, A>::new();
+            tendril.extend_with_byte(BUFFER_SIZE, 0);
             loop {
                 match r.read(&mut tendril) {
                     Ok(0) => return Ok(self.finish()),
                     Ok(n) => {
                         tendril.pop_back(BUFFER_SIZE - n as u32);
-                        self.process(tendril);
+                        self.process(unsafe { tendril.reinterpret_without_validating() });
```
matteoldani, line 87: @simonwuelker let me know if this is acceptable or you have a better idea, open to suggestions

simonwuelker: I think this is okay due to `F: fmt::SliceFormat<Slice = [u8]>`. But really the more I look at this method the more I question its existence (:

simonwuelker (approved):
We can merge this once the formatting errors are addressed.

matteoldani:
I know that a release was just cut for tendril (https://github.com/servo/html5ever/pull/761), however, would it be possible to cut a release with these fixes as well? If not, when should I expect it?

simonwuelker:
The next bump will likely happen when https://github.com/servo/html5ever/pull/750 merges. But if you need these fixes urgently then we can also release a new version immediately.

matteoldani:
Assuming that happens reasonably soon, I have no problem waiting for that. Thanks. 

Meanwhile, I've fixed the linter issue. I think this is ready to be merged

PR #765, opened 2026-07-26
html5ever: simplify `Tokenizer<Sink>::finish_attribute`

niklak:
- `LocalName` compares directly (without pointers)
- `self.current_attr_name` clears after `LocalName` is built, since it's common logic.

modified:   html5ever/src/tokenizer/mod.rs


simonwuelker (approved):
Thanks. There are lots of little cleanups like this that we should do eventually.

PR #766, opened 2026-08-05
html5ever: avoid unnecessary attribute clone in `TreeBuilder::insert_element`.

niklak:
Reordered code in `impl TreeBuilder<Handle, Sink>::insert_element` to avoid an unnecessary clone of tag attributes.

### DIFF
--- a/html5ever/src/tree_builder/mod.rs
+++ b/html5ever/src/tree_builder/mod.rs
@@ -1380,6 +1380,7 @@ where
                 ref prev_element,
             } => (element.clone(), Some(prev_element.clone())),
         };
+
         // Step 12.
         let qname = QualName::new(None, ns, name);
         let form_is_associatable = form_associatable(qname.expanded())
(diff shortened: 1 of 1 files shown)


html5ever/src/tree_builder/mod.rs
```diff
@@ -1388,16 +1380,20 @@ where
...
             && self.form_elem.borrow().is_some()
             && !self.in_html_elem_named(local_name!("template"))
             && !(listed(qname.expanded())
                 && attrs
                     .iter()
-                    .any(|a| a.name.expanded() == expanded_name!("", "form")))
-        {
+                    .any(|a| a.name.expanded() == expanded_name!("", "form")));
+
+        // check the form is associatable, then create element, and we are avoiding of cloning attributes.
```
mrobinson:
```suggestion
        // By checking whether the form is associatable first, then creating the element
        // we can avoid cloning the attributes.
```

html5ever/src/tree_builder/mod.rs
```diff
@@ -1388,16 +1380,20 @@ where
                 ref prev_element,
             } => (element.clone(), Some(prev_element.clone())),
         };
-
```
mrobinson, line 1391: Let's keep this new line.

mrobinson (approved):
Looks good with some stylistic nits:

mrobinson (approved):

PR #767, opened 2026-08-06
Update `actions/checkout` version and use `allow-unsafe-pr-checkout`

mrobinson:


jdm (approved):

.github/workflows/semver-check.yml
```diff
@@ -11,8 +11,9 @@ jobs:
       pull-requests: write # To add labels to PR
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
         with:
+          allow-unsafe-pr-checkout: true
```
jschwe, line 16: I'll try to post a follow-up so we can remove this again.

PR #586, opened 2025-03-23
Store xml namespace attributes for later serialization

devongovett:
Fixes #569

This is @jdm's `xmlns-attrs` branch posted in #569, plus an update to `xml5lib-tests` by @Ygg01 (https://github.com/servo/html5ever/issues/569#issuecomment-2661417199)

PR #772, opened 2026-08-17
Preserve xmlns attributes and rewrite the XML serialization algorithm

### LINKED ISSUE #569: xmlns-prefixed attribute on root element is lost
```xml
<!-- before -->
<svg xmlns="http://www.w3.org/2000/svg" xmlns:sodipodi="http://inkscape.sourceforge.net/DTD/sodipodi-0.dtd">
    <sodipodi:namedview>
        ...
    </sodipodi:namedview>

    <path d="..." sodipodi:nodetypes="cccc"/>
</svg>

<!-- after: xmlns:sodipodi moved to child -->
<svg xmlns="http://www.w3.org/2000/svg">
    <sodipodi:namedview xmlns:sodipodi="http://inkscape.sourceforge.net/DTD/sodipodi-0.dtd">
        ...
    </sodipodi:namedview>

    <path d="..." sodipodi:nodetypes="cccc"></path>
</svg>
```

```xml
<!-- before -->
<svg xmlns="http://www.w3.org/2000/svg" xmlns:sodipodi="http://inkscape.sourceforge.net/DTD/sodipodi-0.dtd">
    <path d="..." sodipodi:nodetypes="cccc"/>
</svg>

<!-- after: xmlns:sodipodi removed entirely -->
<svg xmlns="http://www.w3.org/2000/svg">
    <path d="..." sodipodi:nodetypes="cccc"></path>
</svg>
```

_Originally posted by @noahbald in https://github.com/servo/html5ever/issues/538#issuecomment-2506784772_

Ms2ger:
In #586, it turned out that a targeted change to preserve xmlns attributes exposed issues in the XML serializer that caused test failures in Servo.
This change brings the XML serializer in line with the specification and mostly with other browsers.

The XML serializer needs to keep track of state related to namespaces and their prefixes.
Implementing this correctly is not workable using a unified trait between HTML and XML.
This is a breaking change: to serialize to XML, users now need to use the new xml5ever::serialize::XmlSerializer.

This implements the latest [DOM Parsing and Serialization specification](https://w3c.github.io/DOM-Parsing/).

Fixes #569.

Co-authored-by: Josh Matthews <josh@joshmatthews.net>
Co-authored-by: Devon Govett <devongovett@gmail.com>

### DIFF
--- a/xml5ever/src/lib.rs
+++ b/xml5ever/src/lib.rs
@@ -33,6 +33,9 @@
 #![allow(unexpected_cfgs)]
 #![deny(missing_docs)]
 
+#[macro_use]
+extern crate markup5ever;
+
 pub use markup5ever::*;
 
 pub(crate) mod macros;
--- a/xml5ever/src/lib.rs
+++ b/xml5ever/src/lib.rs
@@ -33,9 +33,6 @@
 #![allow(unexpected_cfgs)]
 #![deny(missing_docs)]
 
-#[macro_use]
-extern crate markup5ever;
-
 pub use markup5ever::*;
 
 pub(crate) mod macros;
--- a/xml5ever/src/serialize/mod.rs
+++ b/xml5ever/src/serialize/mod.rs
@@ -47,13 +47,16 @@ pub struct NamespacePrefixMap {
     scope: BTreeMap<Namespace, Vec<Prefix>>,
 }
 
-impl NamespacePrefixMap {
+impl Default for NamespacePrefixMap {
     /// Create a new `NamespacePrefixMap`, containing only the "xml" prefix for the XML namespace.
-    pub fn new() -> NamespacePrefixMap {
+    fn default() -> Self {
         let mut scope = BTreeMap::new();
         scope.insert(ns!(xml), vec![namespace_prefix!(xml)]);
         NamespacePrefixMap { scope }
     }
+}
+
+impl NamespacePrefixMap {
     fn find(&self, ns: &Namespace, prefix: &Prefix) -> bool {
         let Some(list) = self.scope.get(ns) else {
             return false;
@@ -62,13 +65,11 @@ impl NamespacePrefixMap {
     }
 
     fn add(&mut self, ns: Namespace, prefix: Prefix) {
-        self.scope.entry(ns).or_insert_with(Vec::new).push(prefix);
+        self.scope.entry(ns).or_default().push(prefix);
     }
 
     fn retrieve(&self, ns: &Namespace, preferred: &Option<Prefix>) -> Option<Prefix> {
-        let Some(candidates) = self.scope.get(&ns) else {
-            return None;
-        };
+        let candidates = self.scope.get(ns)?;
         if let Some(preferred) = preferred {
             if candidates.contains(preferred) {
                 return Some(preferred.clone());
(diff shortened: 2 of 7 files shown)


Ms2ger:
Depends on https://github.com/Ygg01/xml5lib-tests/pull/2

xml5ever/src/serialize/mod.rs
```diff
@@ -27,186 +30,459 @@ impl Default for SerializeOpts {
...
-/// parses can read.
+/// parsers can read.
 ///
 /// Serializer contains a set of functions (start_elem, end_elem...)
-/// that make parsing nodes easier.
+/// that make serializing nodes easier.
 pub struct XmlSerializer<Wr> {
     writer: Wr,
-    namespace_stack: NamespaceMapStack,
+    prefix_index: u32,
```
simonwuelker: Please add some brief documentation to this field.

xml5ever/src/serialize/mod.rs
```diff
@@ -27,186 +30,459 @@ impl Default for SerializeOpts {
...
 
-impl NamespaceMapStack {
-    fn new() -> NamespaceMapStack {
-        NamespaceMapStack(vec![])
+impl NamespacePrefixMap {
+    fn find(&self, ns: &Namespace, prefix: &Prefix) -> bool {
+        let Some(list) = self.scope.get(ns) else {
+            return false;
+        };
+        list.contains(prefix)
```
simonwuelker: This would be slightly simpler when written like this:
```rust
self.scope.get(ns).is_some_and(|list| list.contains(prefix))
```

xml5ever/src/serialize/mod.rs
```diff
@@ -27,186 +30,459 @@ impl Default for SerializeOpts {
...
+        let mut local_prefixes_map: BTreeMap<Prefix, Namespace> = BTreeMap::new();
+        // Step 1.
+        let mut local_default_namespace: Option<Namespace> = None;
+        // Step 2.
+        for (name, value) in attributes {
+            // Step 2.1.
+            let attribute_namespace = &name.ns;
+            // Step 2.2.
+            let attribute_prefix = &name.prefix;
+            // Step 2.3.
```
simonwuelker: Please link to the part of the specification that this is implementing and add the step text after the step numbers.

xml5ever/src/lib.rs
```diff
@@ -33,6 +33,9 @@
 #![allow(unexpected_cfgs)]
 #![deny(missing_docs)]
 
+#[macro_use]
+extern crate markup5ever;
```
SimonSapin: Instead of this you should be able to import `use markup5ever::some_macro;` in modules where you use it

Cargo.toml
```diff
@@ -21,9 +21,9 @@ rust-version = "1.85"
 # Repo dependencies
 tendril = { version = "0.5", path = "tendril" }
 web_atoms = { version = "0.3.0", path = "web_atoms" }
-markup5ever = { version = "0.40", path = "markup5ever" }
-xml5ever = { version = "0.40", path = "xml5ever" }
-html5ever = { version = "0.40", path = "html5ever" }
+markup5ever = { version = "0.41", path = "markup5ever" }
+xml5ever = { version = "0.41", path = "xml5ever" }
+html5ever = { version = "0.41", path = "html5ever" }
```
SimonSapin: 0.40.0 is not released yet, so this isn’t necessary if we merge this PR before it is

SimonSapin:
https://github.com/servo/string-cache/pull/308 should fix the division by zero errors on CI

SimonSapin:
https://crates.io/crates/string_cache_codegen/0.11.1 has that fix

PR #779, opened 2026-08-23
xml5ever: keep question marks in processing instruction data

### LINKED ISSUE #774: XML tokenizer drops question marks from processing-instruction data
The tokenizer removes every `?` character from the data of a processing instruction, except the one that forms the closing `?>`. A processing instruction whose data legitimately contains a question mark, such as a stylesheet href with a query string, comes out of the tokenizer with the character missing.

Reproduced on xml5ever 0.39.0 from crates.io, default features.

```toml
xml5ever = "0.39.0"
markup5ever = "0.39.0"
```

## Reproducer

```rust
use std::cell::RefCell;
use markup5ever::buffer_queue::BufferQueue;
use xml5ever::tendril::StrTendril;
use xml5ever::tokenizer::{Pi, ProcessResult, Token, TokenSink, XmlTokenizer};

struct Collector {
    data: RefCell<Vec<String>>,
}
impl TokenSink for Collector {
    type Handle = ();
    fn process_token(&self, token: Token) -> ProcessResult<()> {
        if let Token::ProcessingInstruction(Pi { data, .. }) = token {
            self.data.borrow_mut().push(data.to_string());
        }
        ProcessResult::Continue
    }
}

fn main() {
    let input = r#"<?xml-stylesheet href="style.xsl?v=2"?>"#;
    let sink = Collector { data: RefCell::new(Vec::new()) };
    let bq = BufferQueue::default();
    bq.push_back(StrTendril::from(input));
    let tok = XmlTokenizer::new(sink, Default::default());
    let _ = tok.feed(&bq);
    tok.end();
    println!("{:?}", tok.sink.data.into_inner());
}
```

## Observed

```text
["href=\"style.xslv=2\""]
```

The `?` between `style.xsl` and `v=2` is gone. A standalone input `<?target a?b?c?>` yields data `abc`, with all three interior question marks removed.

## Expected

The data is `href="style.xsl?v=2"`, preserving the question mark. In XML, a processing instruction ends at the first `?>` sequence, so a `?` that is not followed by `>` is ordinary data and belongs in the data string. Both libxml2 and Python `xml.dom.minidom` keep the question mark here.

## Scope

Processing instructions carrying a query string or any interior `?` are silently corrupted. The `xml-stylesheet` instruction with a versioned href is a common real-world case, and the truncated href points at the wrong resource. Any consumer that relies on processing-instruction data reads altered content with no error raised.

## Root cause

The tokenizer reads processing-instruction data one character at a time. It consumes a `?` and inspects the following character. If the following character is not `>`, the code continues reading and does not return the consumed `?` to the data buffer. Only the terminating `?>` needs special handling, and a `?` in any other position belongs in the data.

FadeHack:
Fixes #774

The PI after state was throwing away the `?` that got it there whenever that `?` turned out not to be the start of a `?>`. So this:

```xml
<?xml-stylesheet href="style.xsl?v=2"?>
```

came out of the tokenizer as `href="style.xslv=2"`, pointing at the wrong resource, and nothing was reported. `<?target a?b?c?>` came out as `abc`.

XML 1.0 section 2.6 says a processing instruction runs up to the first `?>`, so a `?` anywhere else is ordinary data and belongs in the data string. libxml2 and Python's minidom both keep it.

### What is in here

Three things, all in the same state:

1. On `?` we now push the pending `?` into the data and stay in the PI after state, so a run of question marks is preserved except for the one that actually closes the instruction.
2. On anything else we push the pending `?` and the character, then go back to the PI data state. It used to push the character and stay in the PI after state, which meant the next `>` closed the instruction even though the character before it was not a `?`. That made `<?target a?b>c?>` stop early with data `ab`. Going back to the PI data state is what the XML5 draft says to do here anyway.
3. At EOF the pending `?` was dropped for the same reason, so an unterminated instruction lost its last character. It is kept now.

### About the skipped test

One xml5lib test, `PI tag with char in PiAfter state`, expects `<?xml \t\n ?m?>` to produce data `m`, so it disagrees with this change. It matches the XML5 draft, whose PI after state appends the `?` when it sees another `?` but silently drops it for anything else. That asymmetry reads like an oversight in the draft rather than something intended, and following it means the data of a processing instruction can silently differ from what every other XML parser produces.

Since the test data lives in a submodule I cannot change here, I skipped that one case with a comment pointing at the issue. Happy to take this a different way if you would rather the draft won, or if you would rather I send a patch to Ygg01/xml5lib-tests first and bump the submodule.

### Testing

Added unit tests in `xml5ever/src/tokenizer/mod.rs` covering the href case from the issue, interior question marks, runs of question marks, the early `>` termination, empty data, and the unterminated case. All three fail without the change.

`cargo test --all`, `cargo fmt --all -- --check` and `cargo clippy --all-features --all-targets` are clean.

### DIFF
--- a/rcdom/tests/xml-tokenizer.rs
+++ b/rcdom/tests/xml-tokenizer.rs
@@ -283,10 +283,26 @@ fn json_to_tokens(js: &Value, exact_errors: bool) -> Vec<Token> {
     sink.get_tokens()
 }
 
-fn mk_xml_test(name: String, input: String, expect: Value, opts: XmlTokenizerOpts) -> Test {
+/// Descriptions of xml5lib tests that we do not run.
+///
+/// The XML5 draft's "PI after state" throws away a '?' that turns out not to be
+/// the start of a '?>', instead of treating it as ordinary data. XML 1.0 §2.6
+/// says a processing instruction runs up to the first '?>', so such a '?'
+/// belongs in the data, and that is what libxml2 and other parsers produce.
+/// This test bakes in the draft's behaviour, so it disagrees with us now that
+/// we keep the character. See https://github.com/servo/html5ever/issues/774
+const IGNORED_TESTS: &[&str] = &["PI tag with char in PiAfter state"];
+
+fn mk_xml_test(
+    name: String,
+    skip: bool,
+    input: String,
+    expect: Value,
+    opts: XmlTokenizerOpts,
+) -> Test {
     Test {
         name,
-        skip: false,
+        skip,
         test: Box::new(move || {
             // Split up the input at different points to test incremental tokenization.
             let insplits = splits(&input, 3);
@@ -308,7 +324,9 @@ fn mk_xml_test(name: String, input: String, expect: Value, opts: XmlTokenizerOpt
 fn mk_xml_tests(tests: &mut Vec<Test>, filename: &str, js: &Value) {
     let input: &str = &js.find("input").get_str();
     let expect = js.find("output");
-    let desc = format!("tok: {}: {}", filename, js.find("description").get_str());
+    let description = js.find("description").get_str();
+    let skip = IGNORED_TESTS.contains(&&*description);
+    let desc = format!("tok: {filename}: {description}");
 
     // Some tests want to start in a state other than Data.
     let state_overrides = vec![None];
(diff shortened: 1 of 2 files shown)


simonwuelker (approved):
Thank you!

PR #781, opened 2026-08-24
Run Miri on CI, make tendril compatible with strict provenance

SimonSapin:
Fixes https://github.com/servo/tendril/pull/64, which the first commit is based on.


tendril/Cargo.toml
```diff
@@ -22,12 +22,3 @@ new_debug_unreachable = { workspace = true }
 
 [dev-dependencies]
 rand = { workspace = true }
-criterion = { workspace = true }
```
SimonSapin, line 25: As dev-dependencies cannot be optional, this allows running Tendril tests under "cross-compiling" Miri without building criterion which depends on `alloca` which requires a C toolchain

mrobinson (approved):

PR #782, opened 2026-08-27
html5ever: Add spec references for the tokenizer steps

simonwuelker:
Our tokenizer code mostly follows the spec directly, but sometimes we deviate. For example, when the specification says to reconsume a character in another state then sometimes we will inline what that state would do, to avoid one iteration of the state machine. This change adds spec text to the code like we do in the `script` crate in servo, and adds notes when our code deviates.

I've also tried to reorder the match arms of the state machine to match the order that they are listed in the spec, but parts of the code depend on the order of the arms, so its not always possible.


mrobinson (approved):
Rubber-stamping this one. It seems like an improvement as long as tests pass.

PR #783, opened 2026-09-12
fix(rcdom): inspect descendant node data in get_a_selects_enabled_selectedcontent (closes #776)

### LINKED ISSUE #776: rcdom: selectedcontent lookup reads self.data, option clone never runs
**Observed**

Parsing HTML with a customizable select leaves the `<selectedcontent>` element empty. `Node::get_a_selects_enabled_selectedcontent` in `rcdom/lib.rs` walks the descendants of the select, but the pattern match inside the loop reads `&self.data` instead of `&node.data` (line 205 on main, line 213 in the published markup5ever_rcdom 0.39.0+unofficial). `self` is the select element itself, so the `local_name!("selectedcontent")` comparison fails on every iteration and the function always returns `None`. As a result the `maybe_clone_an_option_into_selectedcontent` callback from #719 never clones anything, whether it is invoked by the html5ever tree builder or called directly on a correctly shaped tree.

**Expected**

Per https://html.spec.whatwg.org/#clone-an-option-into-a-selectedcontent, closing a selected option inside a select that holds an enabled selectedcontent should replace the children of that selectedcontent with clones of the option's children. Changing the match to `&node.data` produces exactly that tree.

**Reproducer**

```console
cargo new sc-repro && cd sc-repro
cargo add html5ever@0.39 markup5ever_rcdom@0.39
```

`src/main.rs`:

```rust
use html5ever::tendril::TendrilSink;
use html5ever::{parse_document, serialize};
use markup5ever_rcdom::{Handle, NodeData, RcDom, SerializableHandle};

fn find(h: &Handle, name: &str, out: &mut Vec<Handle>) {
    if let NodeData::Element { name: n, .. } = &h.data {
        if &*n.local == name {
            out.push(h.clone());
        }
    }
    for child in h.children.borrow().iter() {
        find(child, name, out);
    }
}

fn main() {
    let input = "<select><button><selectedcontent></selectedcontent></button>\
                 <option selected>Hello</option></select>";
    let dom = parse_document(RcDom::default(), Default::default()).one(input);
    let mut buf = Vec::new();
    let doc: SerializableHandle = dom.document.clone().into();
    serialize(&mut buf, &doc, Default::default()).unwrap();
    println!("serialized: {}", String::from_utf8(buf).unwrap());
    let mut sc = Vec::new();
    find(&dom.document, "selectedcontent", &mut sc);
    println!("selectedcontent children: {}", sc[0].children.borrow().len());
}
```

Output with 0.39.0:

```text
serialized: <html><head></head><body><select><button><selectedcontent></selectedcontent></button><option selected="">Hello</option></select></body></html>
selectedcontent children: 0
```

With the one word fix (`&node.data` on the line noted above) the same program prints `<selectedcontent>Hello</selectedcontent>` and a child count of 1.

**Scope**

Every code path through `maybe_clone_an_option_into_selectedcontent` returns without touching the tree, so the callback introduced in #719 never runs its clone step. Any DOM built with markup5ever_rcdom 0.39.0, by parsing or by direct TreeSink calls, leaves each `<selectedcontent>` element with whatever children it had before the option closed.

rushikeshgarad2024-dev:
### Summary of Fix

Resolves **#776**: `Node::get_a_selects_enabled_selectedcontent` in `rcdom/lib.rs` walked descendant nodes but erroneously matched `&self.data` instead of `&node.data`.

#### Details
- `self` refers to the parent `<select>` element, so `name.local_name() == &local_name!("selectedcontent")` evaluated to false on every iteration, always returning `None`.
- Consequently, the `clone_an_option_into_selectedcontent` routine never ran when parsing customizable selects.
- Changing the pattern match from `&self.data` to `&node.data` allows `<selectedcontent>` descendants to be discovered and properly populated with cloned `<option>` children per WHATWG HTML spec.

Closes #776.

### DIFF
--- a/rcdom/lib.rs
+++ b/rcdom/lib.rs
@@ -202,7 +202,7 @@ impl Node {
         while let Some(node) = remaining.pop_front() {
             remaining.extend(node.children.borrow().iter().cloned());
 
-            let NodeData::Element { name, .. } = &self.data else {
+            let NodeData::Element { name, .. } = &node.data else {
                 continue;
             };
             if name.local_name() == &local_name!("selectedcontent") {


Ygg01:
Shouldn't there be a test to ensure a regression is caught?

rushikeshgarad2024-dev:
Hi @Ygg01! Added a regression unit test `test_get_a_selects_enabled_selectedcontent_descendant` in `rcdom/lib.rs` that explicitly validates finding nested `<selectedcontent>` elements under intermediate elements (e.g. `<select> -> <div> -> <selectedcontent>`) to prevent any regressions. Ready for review! 🚀

